PR #20, opened 2018-01-23
Update syn to 0.12 and quote to 0.4

Eijebong:


Eijebong:
I didn't rename `Data` nor `VariantData` to match syn's new way of naming stuff but I can do that if you want to (this is already a breaking change anyway)

TedDriggs:
Wow, thank you so much for contributing all this. My goal with `Body` and `VariantData` is to match the types exported from `syn`, so if you don't mind switching them I'll start writing up the release notes.

I'm also going to try aliasing the old names to the new types and see if things work; that would at least give library consumers a chance to get warnings before things break.

Eijebong:
There, I also added both type aliases you wanted to add

PR #24, opened 2018-03-04
Support proc-macro2/nightly

### LINKED ISSUE #23: Enabling proc-macro2/nightly breaks this crate
```
error[E0433]: failed to resolve. Could not find `darling` in `{{root}}`
 --> macros\ast_node\src\spanned.rs:3:17
  |
3 | #[derive(Debug, FromMetaItem)]
  |                 ^^^^^^^^^^^^ Could not find `darling` in `{{root}}`
```

This is a known issue. https://github.com/alexcrichton/proc-macro2/issues/67

`quote!` use `Span::def_site()` by default, which is internally `Span::call_site()` on stable.

kdy1:
Fixes #23.

`::darling` style paths only works with `Span::call_site()`.

Note that tests fails with `proc_macro::__internal::with_sess() called before set_parse_sess()!` if `proc- macro2/nightly` is enabled because `proc_macro`'s apis cannot be called from non-proc-macro code.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -17,4 +17,5 @@ default = ["syn/full"]
 [dependencies]
 ident_case = "1.0.0"
 syn = { version = "0.12.10", features = ["extra-traits"] }
+proc-macro2 = "0.2"
 quote = "0.4"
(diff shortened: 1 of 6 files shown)


TedDriggs:
> `::darling` style paths only works with `Span::call_site()`.

I'm not fully understanding this; do you mind expanding on it a bit? Thanks for submitting this PR; going to test it locally then will approve and ship an updated crate if all goes well.

kdy1:
extern crate from call_site (including user-written ones) does not work for def_site because macro hygiene doesn't want spans from multiple places to be mixed.

PR #29, opened 2018-04-28
Have From{Derive,Variant}::from_ident not Option.

upsuper:
`From{Derive,Variant}` only have one place each initializes them, and both places use `Some` on `from_ident`, so it seems unnecessary to use `Option<bool>` on them.

### DIFF
--- a/core/src/codegen/from_derive_impl.rs
+++ b/core/src/codegen/from_derive_impl.rs
@@ -14,7 +14,7 @@ pub struct FromDeriveInputImpl<'a> {
     pub base: TraitImpl<'a>,
     pub attr_names: Vec<&'a str>,
     pub forward_attrs: Option<&'a ForwardAttrs>,
-    pub from_ident: Option<bool>,
+    pub from_ident: bool,
     pub supports: Option<&'a Shape>,
 }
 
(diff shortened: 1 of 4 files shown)


TedDriggs:
This is actually deliberate; it allows us to separate out the omission case from the explicit false case. This was something I learned to do working on derive_builder, and I feel pretty strongly about maintaining fidelity about user input for as long as possible.

upsuper:
I don't really get the point. As can be seen from the change, there is no place initializing those fields to `None` at all, so you are not really maintaining anything, given that the value is always `Some(something)`.

TedDriggs:
Hmmm, I’ll take a look. I may have made a mistake in my code.

upsuper:
Actually, is it really possible for people to declare false for that? It seems to me that having `from_ident` listed means it's true, not having it means false. Is there a third state at all?

TedDriggs:
@upsuper I missed your last comment on this. I'm probably being overly-cautious here, but when I was working on rust-derive-builder (which does a lot of struct --> field inheritance) that being diligent about tracking the three states separately made supporting opt-in and opt-out scenarios easier. In this case, I think you're right though: There's no notion of inheritance for either of those.

PR #30, opened 2018-04-30
Implement FromTypeParam

### LINKED ISSUE #28: Add FromTypeParam
`TypeParam` may also have attributes, so it would be good to have a trait for that as well.

upsuper:
This fixes #28.


core/src/codegen/from_type_param.rs
```diff
@@ -0,0 +1,97 @@
...
+use codegen::{ExtractAttribute, OuterFromImpl, TraitImpl};
+use options::ForwardAttrs;
+
+pub struct FromTypeParamImpl<'a> {
+    pub base: TraitImpl<'a>,
+    pub ident: Option<&'a Ident>,
+    pub attrs: Option<&'a Ident>,
+    pub attr_names: Vec<&'a str>,
+    pub forward_attrs: Option<&'a ForwardAttrs>,
+    pub from_ident: bool,
```
TedDriggs: For the reason mentioned on #29, please use `Option<bool>` here.

tests/from_type_param.rs
```diff
@@ -0,0 +1,58 @@
...
+
+fn extract_type(param: &GenericParam) -> &TypeParam {
+    match *param {
+        GenericParam::Type(ref ty) => ty,
+        _ => unreachable!("Not a type param"),
+    }
+}
+
+#[test]
+fn expand_many() {
```
TedDriggs: Can you add a test with a defaulted type please?

TedDriggs:
One more thought: `darling` has some magic fields that make it easy to get a struct from `DeriveInput`. Should we have a way for the `generics` magic field to be populated with things that implement `FromTypeParam`? That might require a `FromGenerics` trait also be added.

upsuper:
That can be in a followup change. Also type param can have magic fields for things like `bounds`. And to implement `FromGenerics`, it would probably be good to also have `FromConstParam` and `FromLifetimeDef`.

Issue #31, opened 2018-05-03
Magic field `generics` should support user-defined types


TedDriggs:
PR #30 by @upsuper added `FromTypeParam`, which enables extraction of attributes on type parameters. Ideally, `darling` would allow a consuming crate to have that extraction done automatically into the master struct implementing `FromDeriveInput`, getting back to the promise of one-line parsing.

- [x] Add `FromGenerics` trait
-  ~`impl<T: FromTypeParam> FromGenerics for Vec<T>`~
- [x] Decide if there needs to be an `ast::Generics` struct
- [x] Decide if `ast::Generics` needs to preserve lifetime and const `GenericParam` values, and if they need to be preserved in order.

@upsuper I'm working on this, but I'd love your input on the decisions listed above.

upsuper:
Ideally it would be good if we would have lifetime and const eventually, so we should probably take them into consider when trying to design stuff here.

For the last point, I don't think we need to preserve the order between different type of params. Currently lifetime must come before type param. Whether const should come after type params doesn't seem to have been decided, but I suspect that would end up being that way. Even if it's not, I don't think it would be a big problem for users.

This would allow us to split generics into different `Vec`s, one for each type of params (rather than needing to have an enum like what `syn` currently does). That probably also indicates `ast::Generics` is probably not worth it, or we may need a generic struct like
```rust
struct<L, T, C> Generics {
    lifetimes: Vec<L>,
    types: Vec<T>,
    consts: Vec<T>,
}
```
It's probably a bit troublesome for people only want some types of params but not others.

Maybe we can just put them into three magic fields, `type_params`, `lifetime_defs`, and `const_params`.

After thinking about `FromGenerics` for a bit, I suspect `impl<T: FromTypeParam> FromGenerics for Vec<T>` may not be a good idea. If we add e.g. `FromLifetimeDef` in the future, we cannot `impl<T: FromLifetimeDef> FromGenerics for Vec<T>` due to conflicting implementations, which would be quite unfortunate.

I'm not sure what can be done for this... Probably we can have `derive(FromTypeParam)` automatically generate a concrete `impl FromGenerics for Vec<SomeType>`?

upsuper:
The `Generics` may not be that bad, though... We can implement `From{TypeParam,LifetimeDef,ConstParam}` for `()` and `!` and users can just use these if they don't want those params.

TedDriggs:
[...]

PR #32, opened 2018-05-09
Implement FromGenerics et al

TedDriggs:
So #31 has proven to be more involved than I anticipated, but I think all this gets it back to a usable state. A caller would now add `generics` in their struct that implements `FromDeriveInput` and would use `darling::ast::Generics` or `darling::util::WithOriginal` to extract the information they want from the generics while preserving the information they want.

- [ ] Update examples to illustrate use of new features
- [x] Write proper tests (it's been a while since I've had to write new tests for `darling` so I need to see how I set it up)

@upsuper if you have a chance to glance over this, that'd be much appreciated; I'm worried I'm going to do something like the blanket impl that will break things down the line.

### DIFF
--- a/core/src/ast/generics.rs
+++ b/core/src/ast/generics.rs
@@ -47,6 +47,22 @@ impl GenericParamExt for syn::GenericParam {
             None
         }
     }
+
+    fn as_lifetime_def(&self) -> Option<&Self::LifetimeDef> {
+        if let syn::GenericParam::Lifetime(ref val) = *self {
+            Some(val)
+        } else {
+            None
+        }
+    }
+
+    fn as_const_param(&self) -> Option<&Self::ConstParam> {
+        if let syn::GenericParam::Const(ref val) = *self {
+            Some(val)
+        } else {
+            None
+        }
+    }
 }
 
 impl GenericParamExt for syn::TypeParam {
@@ -59,6 +75,7 @@ impl GenericParamExt for syn::TypeParam {
     }
 }
 
+/// A mirror of `syn::GenericParam` which is generic over all its contents.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum GenericParam<T = syn::TypeParam, L = syn::LifetimeDef, C = syn::ConstParam> {
     Type(T),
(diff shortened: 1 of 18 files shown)


src/lib.rs
```diff
@@ -59,7 +59,7 @@ extern crate darling_macro;
 pub use darling_macro::*;
 
 #[doc(inline)]
-pub use darling_core::{FromMetaItem, FromDeriveInput, FromField, FromTypeParam, FromVariant};
+pub use darling_core::{FromMetaItem, FromDeriveInput, FromGenerics, FromField, FromTypeParam, FromVariant};
```
TedDriggs: Need to export `FromGenericParam` here.

core/src/ast/generics.rs
```diff
@@ -0,0 +1,177 @@
...
+    type LifetimeDef = ();
+    type ConstParam = ();
+
+    fn as_type_param(&self) -> Option<&Self::TypeParam> {
+        Some(self)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum GenericParam<T = syn::TypeParam, L = syn::LifetimeDef, C = syn::ConstParam> {
```
upsuper: I don't see how is this useful, given that you are already using `syn::GenericParam` in `struct Generics`.

core/src/ast/generics.rs
```diff
@@ -0,0 +1,177 @@
...
+        None
+    }
+}
+
+impl GenericParamExt for syn::GenericParam {
+    type TypeParam = syn::TypeParam;
+    type LifetimeDef = syn::LifetimeDef;
+    type ConstParam = syn::ConstParam;
+
+    fn as_type_param(&self) -> Option<&Self::TypeParam> {
```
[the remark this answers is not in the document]
TedDriggs: Good point. I'll fix that.

core/src/ast/generics.rs
TedDriggs: I've removed the default for `P` in `ast::Generics`. This allows the user to make their own drop-in replacement for `syn::GenericParam` containing whatever they want it to contain.

PR #33, opened 2018-05-10
Add support for darling attrs on DeriveInput generics

TedDriggs:
With Rust 1.26 adding support for attributes on generic params, darling
can now accept attributes on type params to - for example - override the
emitted type param bound for a specific type param. Doing so required:

1. Moving `options::Core.generics` to a new darling-defined type
2. Making `options::Core::start` fallible to account for errors in parsing darling attrs on type params (this has run-on effects)
3. Adding new `codegen::generics` mod to accept `options::Generics`
4. Changing `codegen::outer_from_impl::compute_impl_bounds` to not apply bound to a specific type param for which the `bound` attr is specified.
5. Adding `darling::ast::TypeParamBounds` to avoid diving deeply into `syn` in consuming code.
6. Stripping attrs from generics before codegen, due to dtolnay/sin#422


TedDriggs:
Regrettably, looks like this didn't land in stable today. @upsuper, I'm going to wait to merge this until Travis passes, but I'll push out v0.5.0 now.

PR #35, opened 2018-05-14
Rename FromMetaItem to FromMeta

### LINKED ISSUE #34: Rename FromMetaItem to FromMeta
`syn` renamed `MetaItem` to `Meta` sometime after `darling` first released. Since `darling` is closely aligned with `syn`, we should probably rename the trait so that new users of the crate don't have to deal with this misalignment. However, this is a pretty big change: `FromMetaItem` is the foundational trait of the entire library.

# Options
1. Do a hard rename of the trait. This is simplest to implement, and shouldn't impact any of the dependent crates since we'd bump the version to 0.6.0
1. Add the new trait and deprecate the old trait in v0.5.1, then cut over in v0.6.0. Without a blanket impl, this could split the (nonexistent) ecosystem, but it's hard to figure out which direction the blanket impl should go.

I'm leaning towards option 1, as @udoprog suggested. The directionality of the blanket impl makes it hard to pick a right transition path.

TedDriggs:
* Change trait name
* Change trait method names
* Rename options and codegen structs

This will fix #34


udoprog:
Cool, I hope it pans out well!

PR #40, opened 2018-05-21
Fix #37: Add UsesTypeParam trait

### LINKED ISSUE #37: Create UsesTypeParams trait
`darling` and other crates in the proc-macro space need to determine which type parameters are used by which fields:

1. `darling` should not add a `FromMeta` bound on type parameters which are used only in skipped fields
1. `derive_builder` should not add a `Clone` bound on type parameters which are used in fields where no setter is avaiable
1. `serde` should not emit `Serialize` or `Deserialize` bounds on fields which are skipped in those respective tasks

I vaguely remember @dtolnay mentioning that this was a hard problem for `serde` to solve, so it'd be really nice for `darling` to provide tools and a model that other crates can pick up.

# Prior Art
[`serde`'s implementation](https://github.com/serde-rs/serde/blob/master/serde_derive/src/bound.rs) has a function `with_bound` that looks relevant to this problem.

# Proposal
```rust
pub trait UsesTypeParams {
    /// Search `self` for `sought_types`, returning a set of refs to type parameter idents found
    /// in this syntax element.
    fn uses_type_params<'a>(&self, sought_types: &'a HashSet<Ident>) -> HashSet<&'a Ident>;
}
```

This trait would be implemented for `syn::Data` and all its children.

# Usage
The caller would need to first invoke the trait over all non-skipped fields and variants, then would need to figure out if any of the "used" type parameters reference indirectly use type parameters via bounds or where clauses. All used (reachable) type parameters would then have the bound added by the caller in the emitted code.

TedDriggs:
This adds a new trait for searching through the AST to find references to type parameters.
That trait is then used by darling's own code generation to avoid bounding generics that aren't settable from attributes.

I'm sure I've missed places where type parameters can appear, so this is probably still a WIP.


core/src/type_params.rs
```diff
@@ -0,0 +1,307 @@
...
+        match *self {
+            syn::Data::Struct(ref v) => v.uses_type_params(type_set),
+            syn::Data::Enum(ref v) => v.uses_type_params(type_set),
+            // Do unions support generics?
+            syn::Data::Union(_) => unimplemented!(),
+        }
+    }
+}
+
+uses_type_params!(syn::Variant, fields);
```
TedDriggs: These should probably be alphabetized.

core/src/type_params.rs
```diff
@@ -0,0 +1,307 @@
...
+        // can't use type parameters.
+        // We don't use `is_empty` here because that also considers punctuation.
+        if self.segments.len() == 0 {
+            return Default::default();
+        }
+
+        // A path segment ident can only match if it is not global and it is the first segment
+        // in the path.
+        let ident_hits = if !self.global() {
+            self.segments[0].ident.uses_type_params(type_set)
```
TedDriggs: @dtolnay is this logic correct, or should I only consider the path's `ident` if the length is exactly 1?

core/src/type_params.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+}
+
+uses_type_params!(syn::TypeTuple, elems);
+
+impl UsesTypeParams for syn::Path {
+    fn uses_type_params<'a>(&self, type_set: &'a IdentSet) -> IdentRefSet<'a> {
+        // Not sure if this is even possible, but a path with no segments definitely
+        // can't use type parameters.
+        // We don't use `is_empty` here because that also considers punctuation.
```
dtolnay: It's not possible for a Punctuated to contain only punctuation.

core/src/type_params.rs
dtolnay: If the intention is to use this for trait bounds then this logic is not what you want. For example the struct in your `qself_vec` test case:

```rust
#[derive(SomeTrait)]
struct Foo<T>(<Vec<T> as a::b::Trait>::AssociatedItem);
```

The unit test expects `T` used, but for the purpose of bounds we probably would not want a `T: SomeTrait` bound in this code.

core/src/type_params.rs
TedDriggs: Thanks, I'll go ahead and clean that up to just use `is_empty` then.

core/src/type_params.rs
TedDriggs: Would we never want this field to be "using" the type parameter? I haven't used the syntax enough to have a good sense of when this would be used and if you'd ever need bounds (or type mapping, or something else).

TedDriggs: FWIW, fitzgen/state_machine_future does track `qself` for exactly the reason I thought. I'm trying to come up with an elegant way of making this work for both; I could try rewriting it as a visitor, but I'm hoping it doesn't come to that.

Another option that comes to mind would be some sort of newtype or tuple struct that sits "around" the type being queried and says "I'm searching for uses to make a new struct/enum". For `TypePath` specifically, it'd intervene and say "consider qself", while in other cases it'd be a passthrough.

A third option would be to add a "purpose" enum and then have that be an extra argument to the `uses_type_params` method. It'd make all uses slightly less ergonomic, but it'd probably be the simplest to implement.

TedDriggs: Thought about this some more, and I’m going to add an `options` parameter that includes a `purpose` field. That keeps the system extensible and composable, and won’t have much impact on callers.

PR #44, opened 2018-10-01
Update syn dependency to 0.15

hcpl:
Far fewer API breakages compared to 0.13 -> 0.14!


TedDriggs:
Thanks! The update has now been published as 0.8.0

PR #65, opened 2019-03-13
chore: Exclude unneeded files from crates.io

ignatenkobrain:


TedDriggs:
Thanks for this PR; I hadn't realized this was an option.

Is the goal of excluding these purely for hygiene, or are they having adverse consequences by being included?

ignatenkobrain:
* It makes archive smaller (not critical, but good)
* It makes Linux distributions to not patch it downstream to not include those files (because they are not needed by anything)

TedDriggs:
Thanks for the clarification. I've merged and will publish v0.8.6 as soon as Travis signs off.

Issue #66, opened 2019-03-19
darling::util::Override::as_mut should return Override<&mut T>


TedDriggs:
Right now it accidentally returns `Override<&T>`, which isn't helpful.

Issue #58, opened 2019-02-06
Make darling_core::options and darling_core::codegen crate-private


TedDriggs:
Once #57 is complete, there's no reason for `darling_core` to expose its code generation structs to the outside world. Hiding these will make refactoring easier without fear of breaking changes.

Issue #49, opened 2019-01-22
Move rust min-version to 1.33.0


TedDriggs:
Now that Rust 2018 is out, moving from 1.18 up to 1.33 would allow the codebase to prepare for 2018 Rust.

TedDriggs:
The code now is on 1.31

PR #67, opened 2019-03-19
V0.9 staging

TedDriggs:
Prepare for v0.9

- Bump rustc to 1.33.0: #49 
- Make `darling_core::{codegen, options}` private: #58 
- Fix `Override::as_mut`: #66 
- Add Clippy attributes: Relates to #46 
- Enable suggestions by default

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 language: rust
 
 rust:
-  - 1.18.0
+  - 1.33.0
   - stable
   - beta
   - nightly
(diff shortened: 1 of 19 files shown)


TedDriggs:
I'm wondering if I should defer enabling `suggestions` by default so that 0.9 can avoid the jump to 1.33. That seems like it'd be an issue for a number of crates, and if there's an older version of `strsim` available then it's not necessary.

TedDriggs:
Dropping `strsim` to 0.7.0 works on 1.18, so I'm closing this PR and opening a new one.

PR #72, opened 2019-03-31
Add impl<T: FromMeta> FromMeta for Vec<T>

zth0:


TedDriggs:
Thanks for submitting this. Do you mind sharing a bit about the use-case where you needed the blanket implementation? I'd been avoiding adding one because I couldn't think of a case where order mattered that wasn't covered by [`darling::util::IdentList`](https://docs.rs/darling/0.9.0/darling/util/struct.IdentList.html).

zth0:
I can't remember what the use case for this was so I'm closing this.

PR #73, opened 2019-04-07
chore: Update strsim to 0.9

ignatenkobrain:


TedDriggs:
Unfortunately, `strsim` stopped being compatible with the versions of rust that darling supports, so we aren't able to take that update at this time.

PR #86, opened 2019-12-26
Fix many clippy lint

Luro02:


TedDriggs:
Thanks for going through all these, and for breaking it out into separate commits. I'm a little hesitant to drop support for 1.31 so quickly, though if we're going to jump to 0.11.0 then maybe it's not a bad idea to do it now.

Luro02:
I think it would be better, to just keep this PR open, until version 0.11 is approaching :).
I will not change all those lines back, because it is kinda pointless and a lot of work (according to the compiler 153 lines).

TedDriggs:
I've started cherry-picking some of the commits over so that we can reduce the number of violations, but I think we're blocked on taking all of them due to `Self` not being valid in all cases on 1.31.

PR #91, opened 2020-03-17
Fix attributes example in README.md

tversteeg:
`e.write_errors()` returns a `proc_macro2::TokenStream` which should be converted to a `proc_macro::TokenStream`.


TedDriggs:
Thanks!

PR #97, opened 2020-06-20
Suggestion for fixing deprecation

TakaakiFuruse:
Hi.

When I was reading your code, I found some deprecated lines of code.
I am not sure which Rust versions you are going to support, so this PR is just a suggestion.

Creating a PR would be more comprehensive than creating an issue and pointing out where to fix, I thought. So, please feel free to close this PR as you do on issues, if you think you don't need it.

### DIFF
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -383,7 +383,7 @@ impl StdError for Error {
         &self.kind.description()
     }
 
-    fn cause(&self) -> Option<&StdError> {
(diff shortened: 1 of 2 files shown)


TedDriggs:
Thanks for submitting this. I haven't decided yet how to handle updating to new Rust versions, which is why I've left these as-is. It looks like I need to fix some issues with Travis-CI, but I'm pretty booked this week so I'll have to take a look next week.

PR #113, opened 2021-01-04
Report errors on unsupported attributes

### LINKED ISSUE #96: Invalid attribute specifications being ignored; should cause compile error.
Absolutely excellent crate! I have been able to reduce the amount of code needed in one of my projects by a pretty drastic degree (thedodd/wither#48). Thanks for all of the excellent work here @TedDriggs!

One issue that I'm running into right now is that if a user attempts to specify a value for an enum attribute being derived, if the user does not use the syntax expected by darling, it looks like the value is either:
- skipped and set to a default if `#[darling(default)]` is specified on the corresponding `FDI` field.
- or if no default attr is present, then a `Missing field` error is generated ... which is going to be deceptive to users because in this case they have specified the field, it is just specified in a way that is apparently unexpected by darling (which my not be intuitive to some).

To use the example which @ubolonton gave in #74 for the type `enum Name`, if the end user were to attempt to provide the attribute as a type literal, say `#[pkg(name = Name::Fn)]` or `#[pkg(name = Name::Str("val"))]`, doing so would trigger the behavior I've described above.

The behavior I would expect would be that an error would be generated telling the user that they've specified the value for the attribute incorrectly. Any thoughts on what we should do to fix this issue?

TedDriggs:
`darling` pre-dates Rust support for non-meta attributes.
These increase flexibility for Rust code, but mean that not `darling`
does not parse all valid attributes.

`darling` previously ignored these outright, creating misleading errors.
This commit makes those explicit errors.

Fixes #96


jonasbb (approved):
This creates an error message like `error: Unable to parse attribute: expected literal`. This solves my problem with #96. Thanks for solving this issue.

TedDriggs:
@jonasbb does the error put the span in the right place? The tests in this repo don't have a good way to show me where the span is referencing, which is unfortunate.

core/src/error/mod.rs
```diff
@@ -70,6 +70,21 @@ impl Error {
...
 
+    /// Creates a new error for an attribute that is a single name-value pair, as opposed to the
+    /// expected `Meta::List` form.
+    pub fn name_value_attribute() -> Self {
+        Error::new(ErrorUnsupportedAttribute::NameValue.into())
+    }
+
+    /// Creates a new error for an attribute that `syn` failed to parse into a `Meta`, blocking
+    /// further parsing.
+    pub fn non_meta_attribute(error: &syn::Error) -> Self {
```
TedDriggs, line 81: Further diagnosis of the problem is out of scope for this PR, but `non_meta_attribute` is part of the public API. Instead of making this take `syn::Error`, is it better to have `darling::AttributeExt::parse_meta_list` which is implemented by `syn::Attribute`? That would allow a smaller public API for error creation, which is good, and would allow recovery logic to be baked in later without another minor version bump (also good).

jonasbb:
@TedDriggs Yes, the span looks right

```
error: Unable to parse attribute: expected literal
  --> $DIR/serde_as.rs:67:55
   |
67 |     #[serde_as(serialize_as = "123", deserialize_as = DisplayFromStr)]
   |                                                       ^^^^^^^^^^^^^^
```

core/src/error/mod.rs
```diff
@@ -70,6 +70,21 @@ impl Error {
         Error::duplicate_field(&path_to_string(path))
     }
 
+    /// Creates a new error for an attribute that is a single name-value pair, as opposed to the
+    /// expected `Meta::List` form.
+    pub fn name_value_attribute() -> Self {
```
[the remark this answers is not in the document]
TedDriggs: This has been solved by moving the parsing of attributes into meta-lists inside `darling::util`, so that better diagnostics do not require future changes to the crate's public API.

core/src/error/mod.rs
TedDriggs: This change has been made.

Issue #110, opened 2020-12-30
Unsure if with_span within FromMeta is working


chipsenkbeil:
I'm porting my custom parsing of attributes over to darling and one such case is marking a field in the form:

```rust
struct MyEnt {
    #[ent(field)]
    field1: u32,

    #[ent(field(indexed))]
    field2: u32,

    #[ent(field(indexed, mutable))]
    field3: u32,

    #[ent(field(mutable))]
    field4: u32,
}
```

### FromMeta derive attempt

Originally, I had this implementation:

```rust
#[derive(FromDeriveInput)]
struct Ent {
    // ...
    data: ast::Data<(), EntField>,
}

#[derive(FromField)]
#[darling(attributes(ent))]
struct EntField {
    // ...
    #[darling(default, rename = "field")]
    attr: Option<FieldAttr>,
}

#[derive(FromMeta)]
struct FieldAttr {
    indexed: bool,
    mutable: bool,
}
```

### Manual implementation

But darling wouldn't support a standalone word of `#[ent(field)]`, so I needed to implement it myself and check for each of the possible attributes:

<img width="996" alt="Screen Shot 2020-12-30 at 3 04 32 PM" src="https://user-images.githubusercontent.com/2481802/103381351-21b35600-4ab1-11eb-8bfa-9686aa27c94f.png">

### Not pointing to location of error

I'm using trybuild to compare my old output to the new before saving it. For the example below, I can't seem to get it to be more specific like I used to have.

<img width="404" alt="Screen Shot 2020-12-30 at 3 14 34 PM" src="https://user-images.githubusercontent.com/2481802/103381557-c0d84d80-4ab1-11eb-8393-1e4a334d24de.png">

<img width="677" alt="Screen Shot 2020-12-30 at 3 11 20 PM" src="https://user-images.githubusercontent.com/2481802/103381417-50313100-4ab1-11eb-8fe8-0f51fa6e9a58.png">

TedDriggs:
I’m traveling today but will take a look tomorrow. Do you have a public repo I can look at? If not, no worries

chipsenkbeil:
I just landed the unfinished migration to darling so you can take a look. This is the file that has some of the code above: https://github.com/chipsenkbeil/entity-rs/blob/main/entity_macros/src/derive/ent/struct/data/internal.rs

I've been piecing together how to use parts of darling from reported issues and reading the source, so I may be approaching some of it incorrectly.

TedDriggs:
[...]

PR #114, opened 2021-01-04
NFC: Provide best practices for errors

TedDriggs:
This improves the documentation about how best to use `darling::Error`

### DIFF
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -54,7 +54,6 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 ///    to ensure those errors appear in the right place. Use `darling::util::SpannedValue` to keep
 ///    span information around on parsed fields so that custom diagnostics can point to the correct
 ///    parts of the input AST.
-/// 4. Enable the `diagnostics` feature of `darling`.
 #[derive(Debug)]
 #[cfg_attr(test, derive(Clone))]
 pub struct Error {
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -26,6 +26,35 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 ///
 /// Given that most errors darling encounters represent code bugs in dependent crates,
 /// the internal structure of the error is deliberately opaque.
+///
+/// # Usage
+/// Proc-macro expansion happens very infrequently compared to runtime tasks such as
+/// deserialization, and it happens in the context of an expensive compilation taks.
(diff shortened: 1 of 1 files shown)


TedDriggs:
@chipsenkbeil can you take a look at this?

TedDriggs:
@chipsenkbeil correct. `darling` errors bind to the tightest span they can, so choosing one of them is likely to point to some random field or meta-item, rather than to any usefully-broad span. Using `Span::call_site()` is the appropriate move for that, though as seen in #110 that will cause the error to point at the macro invocation.

Having looked at `entity-rs`, I don't see it doing anything that precludes the use of `darling::Error` for its custom validations after initial parsing. If there is such a place, can you link me to it? I'd be willing to add a conversion from `syn::Error` into `darling::Error` if you're getting `syn` errors back from the crate as opposed to making them in `entity-rs`.

core/src/error/mod.rs
```diff
@@ -26,6 +26,35 @@ pub type Result<T> = ::std::result::Result<T, Error>;
...
+///    create a local `Vec` to collect errors as they are encountered and then use
+///    `darling::Error::multiple` to create an error containing all those issues if the list
+///    is non-empty after validation. This can create very complex custom validation functions;
+///    in those cases, split independent "validation chains" out into their own functions to
+///    keep the main validator manageable.
+/// 3. Use `darling::Error::custom` to create additional errors as-needed, then call `with_span`
+///    to ensure those errors appear in the right place. Use `darling::util::SpannedValue` to keep
+///    span information around on parsed fields so that custom diagnostics can point to the correct
+///    parts of the input AST.
+/// 4. Enable the `diagnostics` feature of `darling`.
```
TedDriggs, line 57: This still requires nightly, so this isn't a good idea.

```suggestion
```

chipsenkbeil:
> @chipsenkbeil correct. `darling` errors bind to the tightest span they can, so choosing one of them is likely to point to some random field or meta-item, rather than to any usefully-broad span. Using `Span::call_site()` is the appropriate move for that, though as seen in #110 that will cause the error to point at the macro invocation.
> 
> Having looked at `entity-rs`, I don't see it doing anything that precludes the use of `darling::Error` for its custom validations after initial parsing. I did see some places where you use the `?` operator a function that returns a `Result<_, syn::Error>`, so I've added a conversion from `syn::Error` to `darling::Error` to make that seamless in #116

I'm planning to do a little refactoring to clean up the code and split out the derive into multiple derives, so I can work on switching over to `darling::Error` at the same time. Thanks for all of the help and patience!

PR #116, opened 2021-01-05
impl From<syn::Error> for Error

TedDriggs:
`syn` provides useful functions for low-level AST parsing that go beyond
what `darling` attempts to support natively. In these cases, a caller
may want to propagate a `syn::Error` up through a function that returns
a `darling::Result`.

This new impl guarantees lossless handling, emitting the same compiler
error is emitted as if the `syn::Error::to_compile_error` had been
called directly.


TedDriggs:
Merging before CI completion after local testing against 1.31.0 and stable without issue. Filed #117 to address the unacceptable slowness of Travis-CI.

Issue #121, opened 2021-01-29
Add support for Arrays for FromMeta


TurnOfACard:
Support could be added for deriving simple unsigned integer arrays using the FromMeta trait. This provides benefits to the end user as they do not have to perform their own logic to verify validity of integers etc.

For example, a vector could be derived:

```rust
#[derive(derived_array = "[1, 2, 3]")]
pub struct MyStruct {
   field_a: u8,
}

// Would have the corresponding darling code
#[derive(Debug, FromDeriveInput)]
struct DarlingStruct {
    ident: syn::Ident,
    derived_array: Vec<u8>,
}
```

TedDriggs:
This is now published in 0.12.1

PR #122, opened 2021-01-29
FromMeta for unsigned arrays

TurnOfACard:
Addresses issue #121 

Let me know your thoughts.

### DIFF
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -246,7 +246,7 @@ impl<T: syn::parse::Parse, P: syn::parse::Parse> FromMeta for syn::punctuated::P
     }
 }
 
-/// Parsing support for an array, i.e. "[1 + 2, 2 - 2, 3 * 4]".
+/// Parsing support for an array, i.e. "\[1 + 2, 2 - 2, 3 * 4\]".
 impl FromMeta for syn::ExprArray {
     fn from_value(value: &Lit) -> Result<Self> {
         if let Lit::Str(ref ident) = *value {
@@ -259,39 +259,33 @@ impl FromMeta for syn::ExprArray {
     }
 }
 
-macro_rules! from_number_array {
+macro_rules! from_numeric_array {
     ($ty:ident) => {
         impl FromMeta for Vec<$ty> {
             fn from_value(value: &Lit) -> Result<Self> {
-                if let Lit::Str(ref ident) = *value {
-                    let expr_array = ident
-                        .parse::<syn::ExprArray>()
-                        .map_err(|_| Error::unknown_lit_str_value(ident))?;
-                    // To meet rust <1.36 borrow checker rules on expr_array.elems
-                    let v = expr_array
+                let expr_array = syn::ExprArray::from_value(value)?;
+                // To meet rust <1.36 borrow checker rules on expr_array.elems
+                let v =
+                    expr_array
                         .elems
                         .iter()
                         .map(|expr| match expr {
                             Expr::Lit(lit) => $ty::from_value(&lit.lit),
-                            _ => {
-                                Err(Error::unexpected_type(&format!("{:?}", expr)).with_span(expr))
-                            }
+                            _ => Err(Error::custom("Expected array of unsigned integers")
+                                .with_span(expr)),
                         })
                         .collect::<Result<Vec<$ty>>>();
-                    v
-                } else {
-                    Err(Error::unexpected_lit_type(value))
-                }
+                v
             }
         }
     };
 }
 
-from_number_array!(u8);
-from_number_array!(u16);
-from_number_array!(u32);
-from_number_array!(u64);
-from_number_array!(usize);
+from_numeric_array!(u8);
+from_numeric_array!(u16);
+from_numeric_array!(u32);
+from_numeric_array!(u64);
+from_numeric_array!(usize);
 
 /// Parsing support for paths. This attempts to preserve span information when available,
 /// but also supports parsing strings with the call site as the emitted span.
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -267,6 +267,8 @@ macro_rules! from_number_array {
                     let expr_array = ident
                         .parse::<syn::ExprArray>()
                         .map_err(|_| Error::unknown_lit_str_value(ident))?;
+                    // To meet rust <1.36 borrow checker rules on expr_array.elems
+                    let v =
                     expr_array
                         .elems
                         .iter()
@@ -276,7 +278,8 @@ macro_rules! from_number_array {
                                 Err(Error::unexpected_type(&format!("{:?}", expr)).with_span(expr))
                             }
                         })
-                        .collect::<Result<Vec<_>>>()
+                        .collect::<Result<Vec<$ty>>>();
+                    v
                 } else {
                     Err(Error::unexpected_lit_type(value))
                 }
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -272,7 +272,9 @@ macro_rules! from_number_array {
                         .iter()
                         .map(|expr| match expr {
                             Expr::Lit(lit) => $ty::from_value(&lit.lit),
-                            _ => Err(Error::unexpected_type(&format!("{:?}", expr)).with_span(expr)),
+                            _ => {
+                                Err(Error::unexpected_type(&format!("{:?}", expr)).with_span(expr))
+                            }
                         })
                         .collect::<Result<Vec<_>>>()
                 } else {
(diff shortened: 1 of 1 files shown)


core/src/from_meta.rs
```diff
@@ -246,6 +246,53 @@ impl<T: syn::parse::Parse, P: syn::parse::Parse> FromMeta for syn::punctuated::P
...
+                .map_err(|_| Error::unknown_lit_str_value(ident))
+        } else {
+            Err(Error::unexpected_lit_type(value))
+        }
+    }
+}
+
+macro_rules! from_number_array {
+    ($ty:ident) => {
+        impl FromMeta for Vec<$ty> {
```
TedDriggs, line 264: I'm a little concerned about introducing `FromMeta` impls for specific types of `Vec`; wouldn't that prevent a blanket impl in the future?

TurnOfACard: Is a blanket impl using something like `std::str::FromStr` (or other std library traits) already blocked due to darling's implementation of `impl FromMeta for Vec<syn::WherePredicate>`?

TedDriggs: Ah, I think you're right. In that case, I'd like to test how this interacts with `#[darling(multiple)]`, and as long as that works well we should be good.

Can you share a bit more information about the use-case for these?

TurnOfACard: The use case is primarily providing an ergonomic way to provide byte data as an argument to a proc macro. For instance taking slightly modified code from tests/multiple:

`#[hello(ipsum(dolor = 0x12, dolor = 0x34)]
pub struct Baz;
`

It could be rewritten as the following which provides concise code showing the intent of passing an array/vector to the proc macro:
`#[hello(ipsum(dolor = "[0x12, 0x34]")]
pub struct Baz;
`

Whilst it was originally intended for use with a `Vec<u8>`, it was simple to expand via a procedural macro to other unsigned integers so I thought I would. However, mapping the string values from a syn::ExprArray is harder for unsigned integers, as it's returned as an Expr::Unary for negative numbers.

Will you look at interaction with `#[darling(multiple)]` or would you like me to investigate?

core/src/from_meta.rs
```diff
@@ -246,6 +246,53 @@ impl<T: syn::parse::Parse, P: syn::parse::Parse> FromMeta for syn::punctuated::P
...
+            fn from_value(value: &Lit) -> Result<Self> {
+                if let Lit::Str(ref ident) = *value {
+                    let expr_array = ident
+                        .parse::<syn::ExprArray>()
```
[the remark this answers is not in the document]
TurnOfACard: Span is preserved only up to the string it's extracting the value from. As the impl for ExprArray takes it from_value rather than from_string.

For example, providing a value that is not a u8 returns something like:
```
error: Expression did not match expected value
  |
9 |         #[lorem(ipsum = "[0x2, a]")]
  |                         ^^^^^^^^^^
```

However you raise a good point, the return error value is not helpful. Without attempting to handle every `syn::Expr` case which may lead to misleading types (i.e., Unexpected type `syn::Expr::Unary`, or in the case of provision of strings without quotation marks, `syn::Expr::Path`).

In this case I'd think that an error message such as "Expression did not match expected value" or "Expected u8" may be suitable?. I've marked a suggestion above. There is also a need to update the doc comment, as it will throw a warning for the doc comment "[1 + 2, 2 - 2, 3 * 4]".

core/src/from_meta.rs
```diff
@@ -246,6 +246,53 @@ impl<T: syn::parse::Parse, P: syn::parse::Parse> FromMeta for syn::punctuated::P
...
+        } else {
+            Err(Error::unexpected_lit_type(value))
+        }
+    }
+}
+
+macro_rules! from_number_array {
+    ($ty:ident) => {
+        impl FromMeta for Vec<$ty> {
+            fn from_value(value: &Lit) -> Result<Self> {
```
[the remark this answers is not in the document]
TurnOfACard: I would imagine `from_value` preserves span info over `from_string`. If this is not the case let me know.

core/src/from_meta.rs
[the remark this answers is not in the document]
TedDriggs: I think having the error specify that it expected an array of unsigned integers would be great. Having the error span the whole string is fine; frankly I'm not sure it's possible to provide more granularity than that.

core/src/from_meta.rs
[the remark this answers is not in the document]
TedDriggs: The default implementation of `from_value` delegates to `from_string` and then sets the error span correctly. However, you might be right that `syn` would be unhappy in that case.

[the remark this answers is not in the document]
TedDriggs: Yeah, it looks like `syn` does lose span information; see [this example](https://docs.rs/darling_core/0.12.0/src/darling_core/from_meta.rs.html#229).

core/src/from_meta.rs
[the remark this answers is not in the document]
TurnOfACard: When trying to improve the errors, I've noticed that the call `ident.parse::<syn::ExprArray>()` can write an error we cannot catch if input such as `#[lorem(ipsum = "[0x2, a")]` is provided, it returns an error "this file contains an unclosed delimiter".

From what I can tell, it's caused by this call in [proc-macro2](https://github.com/alexcrichton/proc-macro2/blob/666060416a9a9e466e491ac7ba18e334d0dfa6af/src/lib.rs#L183) which leads into the rust core library. I spent some time looking but that is far beyond my knowledge.

It also seems to trigger if you were to derive a syn::Path in darling so `#[lorem(ipsum = "[a")]` where ipsum is a Path triggered the same outcome. I will update with a better error, but happy to discuss the "unclosed delimiter" error.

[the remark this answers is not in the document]
TedDriggs: That seems fine.

core/src/from_meta.rs
```diff
@@ -246,6 +246,47 @@ impl<T: syn::parse::Parse, P: syn::parse::Parse> FromMeta for syn::punctuated::P
     }
 }
 
+/// Parsing support for an array, i.e. "\[1 + 2, 2 - 2, 3 * 4\]".
```
TedDriggs, line 249:
```suggestion
/// Parsing support for an array, i.e. `example = "[1 + 2, 2 - 2, 3 * 4]"`.
```

core/src/from_meta.rs
```diff
@@ -246,6 +246,47 @@ impl<T: syn::parse::Parse, P: syn::parse::Parse> FromMeta for syn::punctuated::P
...
+                .map_err(|_| Error::unknown_lit_str_value(ident))
+        } else {
+            Err(Error::unexpected_lit_type(value))
+        }
+    }
+}
+
+macro_rules! from_numeric_array {
+    ($ty:ident) => {
+        impl FromMeta for Vec<$ty> {
```
TedDriggs, line 264: An example baked into the macro would be good too. It doesn't need to run - illustrating how it can be used like you did for the `ExprArray` would be sufficient.

core/src/from_meta.rs
TedDriggs: I ran this test, which worked:

```rust
#[macro_use]
extern crate darling;
#[macro_use]
extern crate syn;

use darling::FromDeriveInput;
use syn::DeriveInput;

#[derive(Debug, FromDeriveInput)]
#[darling(attributes(lorem))]
struct Receiver {
    // #[darling(multiple)]
    value: Vec<u8>,
}

fn main() {
    let di: DeriveInput = parse_quote! {
        #[derive(Receiver)]
        #[lorem(value = "[1, 2]")]
        // #[lorem(value = "1", value = "2")]
        struct Foo;
    };

    dbg!(Receiver::from_derive_input(&di).unwrap());
}
```

Uncommenting the attribute declaration in `Receiver` and swapping which one is used on `Foo` has the desired effects, so I think we can consider this resolved.

core/src/from_meta.rs
```diff
@@ -246,6 +246,48 @@ impl<T: syn::parse::Parse, P: syn::parse::Parse> FromMeta for syn::punctuated::P
...
+        } else {
+            Err(Error::unexpected_lit_type(value))
+        }
+    }
+}
+
+/// Parsing support for an unsigned integer array, i.e. `example = "[1, 2, 3, 4]"`.
+macro_rules! from_numeric_array {
+    ($ty:ident) => {
+        impl FromMeta for Vec<$ty> {
```
TedDriggs, line 265:
```suggestion
macro_rules! from_numeric_array {
    ($ty:ident) => {
        /// Parse an unsigned integer array, i.e. `example = "[1, 2, 3, 4]"`.
        impl FromMeta for Vec<$ty> {
```

TedDriggs: Unless I'm very much mistaken, putting the comment here will cause it to show up for the emitted impls, which will let someone see the syntax to pass this.

TurnOfACard: No worries, I had thought about that but can't recall why I chose to document the macro over the impl.

PR #126, opened 2021-04-07
Fix FromMeta for chars

okryvyts:
Hello

Yesterday I spent all evening figuring out why my proc macro cannot recognize char attributes after I redefined FromMeta::from_char for my newtype. Here is the fix for the issue. I also add FromMeta impl for char as I don't see any reason why darling doesn't support chars out of the box. 

Nice lib, btw))

### DIFF
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -163,8 +163,12 @@ impl FromMeta for char {
     }
 
     fn from_string(s: &str) -> Result<Self> {
-        if s.chars().count() == 1 {
-            Ok(s.chars().next().unwrap())
+        let mut chars = s.chars();
+        let char1 = chars.next();
+        let char2 = chars.next();
+
+        if let (Some(char), None) = (char1, char2) {
+            Ok(char)
         } else {
             Err(Error::unexpected_type("string"))
         }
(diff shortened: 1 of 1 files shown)


core/src/from_meta.rs
```diff
@@ -152,6 +157,20 @@ impl FromMeta for AtomicBool {
...
+    fn from_char(value: char) -> Result<Self> {
+        Ok(value)
+    }
+
+    fn from_string(s: &str) -> Result<Self> {
+        if s.chars().count() == 1 {
+            Ok(s.chars().next().unwrap())
+        } else {
+            Err(Error::unexpected_type("string"))
+        }
```
Veetaha, line 170: Let's avoid unwrap here since we can
```suggestion
		let mut chars = s.chars();
		let char1 = chars.next();
		let char2 = chars.next();

        if let (Some(char), None) = (char1, char2) {
            Ok(char)
        } else {
            Err(Error::unexpected_type("string"))
        }
```

Veetaha (approved):

TedDriggs:
Thanks! Glad you are enjoying the library. I'll ship an update with this now.

TedDriggs:
v0.12.3 is live with this change.

okryvyts:
Nice, thanks

PR #132, opened 2021-04-21
Another Clippy PR

kangalioo:
There seems to be a PR for this already (#86) plus a clippy branch in this repository. Both of those seem to be kind of abandoned though. Plus you expressed concern in #86 about some of the changes not working in 1.31, which is Darling's minimum supported Rust version.

This PR tries to address those issues by being based on the latest master branch and using a `clippy.toml` file with `msrv = "1.31.0"`. That way, Clippy will only suggest lints relevant to that version.

The most common lint was explicit lifetimes where they could be elided. Apart from that were a few miscellaneous lints, some were false positives.

Should Clippy warnings be checked in GitHub Actions CI?


TedDriggs (approved):
Thanks for doing this!

TedDriggs:
It would be great if CI ran clippy

Issue #117, opened 2021-01-05
Investigate GitHub actions for CI


TedDriggs:
Travis-CI is now taking upwards of 40 minutes to _start_ the CI job, which is not conducive to continuous testing.

chipsenkbeil:
I just went through this in case you want a reference for Github actions: https://github.com/chipsenkbeil/entity-rs/blob/main/.github/workflows/ci.yml

I test the default features, all features, and no default features. Because of the way dependencies are pulled in during all features, I had to limit it to just the entity crate instead of all crates. For no default features, you can't use the workspace option to select everything and instead have to specify a manifest, but I don't think either of these warts would apply to darling.

TedDriggs:
@chipsenkbeil thanks for the guidance! Yours was a great starting point, and the speed improvement over Travis is frankly stunning.

chipsenkbeil:
@TedDriggs, I just discovered that my CI was missing an `override: true` in the test section. You'll need that to get it to use the proper toolchain, which is why my nightly run wasn't detected by rustversion.

![Screen Shot 2021-01-06 at 10 52 14 AM](https://user-images.githubusercontent.com/2481802/103796785-4194cd80-500d-11eb-9e71-6d11f0647f5a.png)

PR #133, opened 2021-04-22
Another Clippy PR part 2

kangalioo:
After #132, this PR introduces Clippy into GitHub Actions CI and also fixes a few lints that I didn't catch in the previous pull request because I forgot to use newest Clippy, turn on all features, and turn on lints for tests (those things are now enabled in CI so hopefully nobody can forget again).

The workflow code is slightly adapted from https://github.com/chipsenkbeil/entity-rs/blob/main/.github/workflows/ci.yml, which is where most of darling's workflow file comes from as far as I can tell (https://github.com/TedDriggs/darling/issues/117#issuecomment-754720411).

The `blacklisted-names = []` line in clippy.toml was added because Clippy linted the usage of `foo` in various integration test files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,12 +64,12 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: stable
           override: true
           components: clippy
       - uses: Swatinem/rust-cache@v1
-      - name: Run clippy --workspace --all-features --tests
-        run: cargo clippy --workspace --all-features --tests
+      - name: Run clippy --workspace --tests
+        run: cargo clippy --workspace --tests
 
   rustfmt:
     name: Verify code formatting
--- a/core/src/error/kind.rs
+++ b/core/src/error/kind.rs
@@ -130,7 +130,7 @@ impl ErrorUnknownField {
     }
 
     #[cfg(feature = "diagnostics")]
-    pub fn to_diagnostic(self, span: Option<::proc_macro2::Span>) -> ::proc_macro::Diagnostic {
+    pub fn into_diagnostic(self, span: Option<::proc_macro2::Span>) -> ::proc_macro::Diagnostic {
         let base = span
             .unwrap_or_else(::proc_macro2::Span::call_site)
             .unwrap()
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,1 +1,2 @@
 msrv = "1.31.0"
+blacklisted-names = [] # we want to be able to use placeholder names in tests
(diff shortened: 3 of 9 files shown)


core/src/error/kind.rs
```diff
@@ -130,7 +130,7 @@ impl ErrorUnknownField {
     }
 
     #[cfg(feature = "diagnostics")]
-    pub fn to_diagnostic(self, span: Option<::proc_macro2::Span>) -> ::proc_macro::Diagnostic {
+    pub fn into_diagnostic(self, span: Option<::proc_macro2::Span>) -> ::proc_macro::Diagnostic {
```
TedDriggs, line 133: Can you please add this to `CHANGELOG.md` under an "Unreleased" header at the top? This is technically a breaking change.

.github/workflows/ci.yml
```diff
@@ -54,6 +54,23 @@ jobs:
...
+    name: Lint with clippy
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: -Dwarnings
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
```
TedDriggs, line 67: Why nightly here? Wouldn't `stable` produce more CI-friendly output?

kangalioo: If Clippy runs with `--all-features`, it seems that nightly must be used or the `diagnostics` feature will not compile.

If you want, Clippy can also run without `--all-features` in stable, but in turn it can't lint against `diagnostics` code.

What do you mean with CI-friendly output, by the way?

core/src/error/kind.rs
kangalioo: Oh is it? I thought this method is not public because the containing struct is marked `pub(in crate::error)` and the function is not shown [in the online docs](https://docs.rs/darling/0.12.4/darling/index.html?search=to_diagnostic)

.github/workflows/ci.yml
TedDriggs: > or the `diagnostics` feature will not compile.

Sadly, I don't see that stabilizing anytime soon in rustc. It's probably more important to lint the case without that feature enabled.

> What do you mean with CI-friendly output, by the way?

I've been burned by nightly having changes that caused builds to fail for "no reason." Forget I said anything 😄

core/src/error/kind.rs
TedDriggs: Right you are.

.github/workflows/ci.yml
kangalioo: Okay, the Clippy install is now changed back to stable and runs on default features

PR #155, opened 2022-01-25
Add Error::ok_if_empty

### LINKED ISSUE #115: Make error collection easier
Today, collecting multiple custom errors requires writing:

```rust
let mut errors = vec![];

//... validation code

if !errors.is_empty() {
    return Err(darling::Error::multiple(errors));
}
```

That's not _bad_, but it's enough boilerplate that it's tempting to write a macro for it.

An alternative would be:

```rust
let mut errors = vec![];

// ... validation

darling::Error::check_none(errors)?;
```

The `check_none` function would return `darling::Result<()>`; it would return `Ok` if the list is empty, and `Err` if there were errors. The try operator counts as using the `Result`, so no `let _ =` is required.

# Open Questions
1. Is this useful enough to merit expanding the API surface?
2. Is this the right API? Alternatives would include one where the caller asks for an "error collector" rather than using a standard `Vec`, or a free function in `darling::error` rather than a function off `darling::Error`.
3. Is there a preexisting name for this concept from other libraries that we should name it here?

TedDriggs:
This slightly reduces typing for the darling error-collection paradigm.

Open question: Is this worth it? I'm not convinced the code is much more readable.

Fixes #115

### DIFF
--- a/core/src/ast/data.rs
+++ b/core/src/ast/data.rs
@@ -130,11 +130,9 @@ impl<V: FromVariant, F: FromField> Data<V, F> {
                     }
                 }
 
-                if !errors.is_empty() {
-                    Err(Error::multiple(errors))
-                } else {
-                    Ok(Data::Enum(items))
-                }
+                Error::ok_if_empty(errors)?;
+
+                Ok(Data::Enum(items))
             }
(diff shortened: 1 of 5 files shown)


TedDriggs:
This has been superseded by a better API in #164

PR #159, opened 2022-03-06
impl FromMeta for syn::Expr

kuviman:
Needed expressions in attributes, so just implemented it

Although now I see I could have just used `map`


TedDriggs:
The code change here looks great; I've added some unit-tests in [this branch](https://github.com/TedDriggs/darling/tree/expr-from_meta) that I think would be good to include.

Issue #161, opened 2022-03-07
`Option` handling is non-intuitive


CraftSpider:
**Expectations**:
Based on the phrasing in the documentation and the fact that the crate says it is inspired by serde, `Option` would default to `None` in a derive if the item isn't present.

**Reality**:
Missing items error, and need an explicit `#[darling(default)]` on them. This seems non-obvious, and at the least I think the documentation for `Option` in `FromMeta` should be clarified, as while the current documentation technically describes the behavior accurately, I misinterpreted it as I assumed it would work the same as `syn`.

TedDriggs:
In `serde`, options aren't automatically defaulting; if you don't set `#[serde(default)]` an explicit `null` would be required in the input for `Deserialize` to work.

> I assumed it would work the same as `syn`.

When I started `darling`, I don't think `syn` had any behavior here. I can see `Option` making something implicitly optional as useful for saving typing, but it does come at the expense of irregularity (since other types wouldn't implicitly be defaulted)

CraftSpider:
Thanks for the response. I just checked on the playground, at least with `serde`/`serde_json` it does automatically default:
```rs
#[derive(Debug, serde::Deserialize)]
struct Foo {
    item: Option<i32>,
}

fn main() {
    dbg!(serde_json::from_str::<Foo>("{}")
        .unwrap());
}
```
This succeeds, printing out `Foo { item: None }`. On the other hand, I can understand not wanting to add the irregularity. I think it would be nice to have a note in the documentation, but feel free to close the issue if you want.

TedDriggs:
Today I learned.

Knowing that, I'm certainly willing to entertain a PR that adds this, though I may sit agonizing about it for a week before merging. A couple things to consider:

1. Determining that a field is an `Option` isn't quite as easy as it sounds. [Here](https://github.com/colin-kiegel/rust-derive-builder/blob/master/derive_builder_core/src/setter.rs#L218) is an example from `derive_builder` - if you can find where `serde` does this check that'd be even better.
2. See example below

```rust
#[derive(FromMeta)]
#[darling(default)]
struct Example {
    name: Option<String>,
    age: u8,
}

impl Default for Example {
    fn default() -> Self {
        Self {
            name: Some("Alice".into()),
            age: 0,
        }
    }
}
```

[...]

PR #163, opened 2022-03-08
WIP: Infer default for Option fields

TedDriggs:
Related to #161

### DIFF
--- a/core/src/codegen/field.rs
+++ b/core/src/codegen/field.rs
@@ -2,10 +2,31 @@ use std::borrow::Cow;
 
 use proc_macro2::TokenStream;
 use quote::{ToTokens, TokenStreamExt};
-use syn::{Ident, Path, Type};
+use syn::{Ident, Path, PathArguments, Type};
 
 use crate::codegen::{DefaultExpression, PostfixTransform};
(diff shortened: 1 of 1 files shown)


core/src/codegen/field.rs
```diff
@@ -7,6 +7,18 @@ use syn::{Ident, Path, Type};
 use crate::codegen::{DefaultExpression, PostfixTransform};
 use crate::usage::{self, IdentRefSet, IdentSet, UsesTypeParams};
 
+fn is_option(ty: &Type) -> bool {
```
TedDriggs, line 10: This is something `derive_builder` and `serde` also do. Should it be extracted to be part of `darling::util`?

core/src/codegen/field.rs
```diff
@@ -173,7 +197,7 @@ impl<'a> ToTokens for Initializer<'a> {
             } else {
                 quote!(#ident: #ident)
             }
-        } else if let Some(ref expr) = field.default_expression {
+        } else if let Some(ref expr) = self.option_aware_default() {
```
TedDriggs, line 200: Should we do something here to make spans better?

PR #164, opened 2022-03-17
Error collector

ijackson:
I noticed the poor ergonomics of `Error::multiple`.  I see there's an issue for this already, #115.

Here is my idea of what a nice API woudl be like.  It handles not only the awkward `if errors.is_empty()` at the end, but also collection of errors as we go.  This is IMO a better alternative to #155.

### DIFF
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -578,7 +578,7 @@ impl Accumulator {
     fn errors(&mut self) -> &mut Vec<Error> {
         match &mut self.0 {
             Some(errors) => errors,
-            None => panic!("darling internal error"),
+            None => panic!("darling internal error: Accumulator accessed after defuse"),
         }
     }
 
@@ -589,7 +589,7 @@ impl Accumulator {
     pub fn into_inner(mut self) -> Vec<Error> {
         match std::mem::replace(&mut self.0, None) {
             Some(errors) => errors,
-            None => panic!("darling internal error"),
+            None => panic!("darling internal error: Accumulator accessed after defuse"),
         }
     }
 
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -504,6 +504,19 @@ impl Iterator for IntoIter {
 ///
 /// See the docs for [`darling::error::Error`](Error) for more discussion of error handling with darling.
 ///
+/// # Panics
+///
+/// `Accumulator` has a *drop bomb*: it panics if it is dropped.
+/// You must call [`finish()`](Accumulator::finish)
+/// (or [`into_inner`](Accumulator::into_inner)).
+///
+/// You must properly `finish` the `Accumulator`,
+/// *even if it has never been used* and does not contain any errors.
+/// If you want to discard an `Accumulator` that you know to be empty,
+/// write something like `accumulator.finish(()).unwrap()`.
+///
+/// # Example
+///
 /// ```
 /// # extern crate darling_core as darling;
 /// # struct Thing;
@@ -526,8 +539,17 @@ impl Iterator for IntoIter {
 /// ```
 #[derive(Default, Debug)]
 #[must_use = "you must properly finish() an Accumulator"]
-pub struct Accumulator {
-    errors: Vec<Error>,
+pub struct Accumulator(AccumulatorInner);
+
+/// Because Drop::drop takes `&mut self`, rather than `self`, etc.
+#[derive(Debug)]
+enum AccumulatorInner {
+    Live(Vec<Error>), // Drop bomb is live
+    Defused, // Exists only after passing into a public method taking owned `self`.
+}
+
+impl Default for AccumulatorInner {
+    fn default() -> Self { AccumulatorInner::Live(vec![]) }
 }
 
 impl Accumulator {
@@ -555,7 +577,7 @@ impl Accumulator {
                 Some(y)
             }
             Err(e) => {
-                self.errors.push(e);
+                self.push(e);
                 None
             }
         }
@@ -568,22 +590,33 @@ impl Accumulator {
     /// If there were no errors recorded, returns `Ok(success)`.
     /// Otherwise calls [`Error::multiple`] and returns the result as an `Err`.
     pub fn finish<T>(self, success: T) -> Result<T> {
-        if self.errors.is_empty() {
+        let errors = self.into_inner();
+        if errors.is_empty() {
             Ok(success)
         } else {
-            Err(Error::multiple(self.errors))
+            Err(Error::multiple(errors))
+        }
+    }
+
+    fn errors(&mut self) -> &mut Vec<Error> {
+        match &mut self.0 {
+            AccumulatorInner::Live(errors) => errors,
+            AccumulatorInner::Defused => panic!("darling internal error"),
         }
     }
 
     /// Returns the collected errors as a plain `Vec`
     #[must_use = "the accumulated errors should not simply be dropped"]
-    pub fn into_inner(self) -> Vec<Error> {
-        self.errors
+    pub fn into_inner(mut self) -> Vec<Error> {
+        match std::mem::replace(&mut self.0, AccumulatorInner::Defused) {
+            AccumulatorInner::Live(errors) => errors,
+            AccumulatorInner::Defused => panic!("darling internal error"),
+        }
     }
 
     /// Add one error
-    pub fn push(&mut self, item: Error) {
-        self.errors.push(item)
+    pub fn push(&mut self, error: Error) {
+        self.errors().push(error)
     }
 
     /// Check if we have collected errors, and if so produce an aggregate error right away
@@ -597,10 +630,20 @@ impl Extend<Error> for Accumulator {
     where
         I: IntoIterator<Item = Error>
     {
-        self.errors.extend(iter)
+        self.errors().extend(iter)
     }
 }
 
+impl Drop for Accumulator {
+    fn drop(&mut self) {
+        match &self.0 {
+            AccumulatorInner::Defused => {}
+            AccumulatorInner::Live(errors) => {
+                panic!("bug in macro using darling: darling::error::Accumulator dropped (with {} errors), rather than passed to finish()", errors.len())
+            }
+        }
+    }
+}
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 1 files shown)


ijackson:
I experimented with this in `derive_builder` on top of my RFC https://github.com/colin-kiegel/rust-derive-builder/pull/243 and the result I came up with looks like this (using the fact that `Option` can be an iterator to pass it to `extend`; other approaches, eg like the one I put here in this MR in the doc comment, are possible):
```
    let mut errors = darling::error::Collector::new();

    for attr in input.drain(..) {
        let unnest = outputs
            .iter_mut()
            .find(|(ptattr, _)| attr.path.is_ident(ptattr));

        if let Some((_, unnest)) = unnest {
            let unnested = errors.ok_of(unnest_from_one_attribute(attr));
            unnest.extend(unnested);
        } else {
            for (_, output) in outputs.iter_mut() {
                output.push(attr.clone());
            }
        }
    }

    errors.conclude(())
```

TedDriggs:
I think having a `Collector` type is reasonable, but I would do the API a bit differently.

```rust
let mut errors = darling::Error::collector();

for attr in input.drain(..) {
    let unnest = outputs
        .iter_mut()
        .find(|(ptattr, _)| attr.path.is_ident(ptattr));

    if let Some((_, unnest)) = unnest {
        match unnest_from_one_attribute(attr) {
            Ok(attr) => unnest.push(attr),
            Err(e) => errors.push(e)
        }
    } else {
        for (_, output) in outputs.iter_mut() {
            output.push(attr.clone());
        }
    }
}

errors.checkpoint()
```

Key differences:

* Have the collector be created off `darling::Error` so someone doesn't need to import more things
* Keep the imperative `errors.push` approach, primarily for drop-in compat with existing code, and because I think `Option` having an `IntoIter` is so clever it's easy to make a mistake with.
* Add `fn checkpoint(&mut self) -> darling::Result<()>`; this could be in addition to `finish`.

My main concern with this is the lack of prior art; I haven't had time to look for other examples of crates that accumulate/collect errors to see how they do it.

I wonder if this is a `Context` or something? That naming might be better.

ijackson:
> Key differences:
> 
>     * Have the collector be created off `darling::Error` so someone doesn't need to import more things

Makes sense to me.  It should probably still `impl Default` in case that's helpful to someone.

>     * Keep the imperative `errors.push` approach, primarily for drop-in compat with existing code, and because I think `Option` having an `IntoIter` is so clever it's easy to make a mistake with.

I think whether using `extend` is nice is a question about the code in `derive_builder` really :-).  I do think it's really worth supporting is some convenience approach that can be used to wrap error-generating code.  What did you think of this bit in my docs example (edited lightly to remove a pedagogic type decatation) ?
```
for thing in inputs {
    errors.run(||{
        let validateed = thing.validate()?;
        outputs.push(validateed);
        Ok(())
    });
}

```

>     * Add `fn checkpoint(&mut self) -> darling::Result<()>`; this could be in addition to `finish`.

That could be useful.

> My main concern with this is the lack of prior art; I haven't had time to look for other examples of crates that accumulate/collect errors to see how they do it.

I did look for some prior art.  I'm sure I've seen *something* like this somewhere but wasn't able to find it.  I don't think there's any in std.

> I wonder if this is a `Context` or something? That naming might be better.

I find `Context` is a very common word which people use for everything.  OK, it will be called `darling::error::Context`, but I still think that's vague.  I think a more precise word is probably better.

ijackson:
Hmm, `mem::take` is too new.  I will fix it.

core/src/error/mod.rs
```diff
@@ -47,9 +47,9 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 ///    This preserves all span information, suggestions, etc. Wrapping a `darling::Error` in
 ///    a custom error enum works as-expected and does not force any loss of fidelity.
 /// 2. Do not use early return (e.g. the `?` operator) for custom validations. Instead,
-///    create a local `Vec` to collect errors as they are encountered and then use
-///    `darling::Error::multiple` to create an error containing all those issues if the list
-///    is non-empty after validation. This can create very complex custom validation functions;
+///    create an [`error::Collector`](Collector) to collect errors as they are encountered.  Then use
```
TedDriggs, line 50: I kind of like the name `Accumulator` or `ErrorAccumulator` for this, given that `collect` has strong meanings to Rust.

TedDriggs:
Does it make sense for `Collector` to have a drop-bomb, so that will panic if it's dropped while not empty? It seems like in most cases, it's a mistake to forget to call `checkpoint` or `conclude`, and a panic seems more useful than errors silently disappearing (and possibly leading to bad code generation).

core/src/error/mod.rs
```diff
@@ -491,6 +500,104 @@ impl Iterator for IntoIter {
...
+    }
+
+    /// Add one error
+    pub fn push(&mut self, item: Error) {
+        self.errors.push(item)
+    }
+
+    /// Check if we have collected errors, and if so produce an aggregate error right away
+    pub fn checkpoint(&mut self) -> Result<()> {
+        std::mem::take(self).conclude(())
```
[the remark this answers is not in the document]
ijackson: I could try it the other way but I suspect it might involve an `unwrap` or equivalent, which this avoids.

core/src/error/mod.rs
```diff
@@ -491,6 +500,104 @@ impl Iterator for IntoIter {
...
+    /// If you need an actual value `T` for further processing even in error cases,
+    /// rather than a `None`,
+    /// use
+    /// [`Option::unwrap_or_default`],
+    /// [`unwrap_or_else`](Option::unwrap_or_else) or
+    /// [`unwrap_or`](Option::unwrap_or)
+    /// on the return value from `ok_of` or `run`.
+    pub fn ok_of<T>(&mut self, result: Result<T>) -> Option<T> {
+        match result {
+            Ok(y) => {
```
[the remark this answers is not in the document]
ijackson: I may not have done.  I will do so...

core/src/error/mod.rs
ijackson: Makes sense.

ijackson:
> Does it make sense for `Collector` to have a drop-bomb, so that will panic if it's dropped while not empty? It seems like in most cases, it's a mistake to forget to call `checkpoint` or `conclude`, and a panic seems more useful than errors silently disappearing (and possibly leading to bad code generation).

Interesting idea.  I think I like it.  But it ought to happen on *any* plain drop, so that ytou can't ship code that triggers it by mistake, even if you only test the success case.  This is slightly fiddly inside `Accumulator` since it has to distinguish "consumed by `finish`" from "simply empty", but I think it's worth it.

I'll revise this MR tomorrow.

TedDriggs:
That's true, though `checkpoint` does need to defuse it so it's also not as simple as defusing only once.

ijackson:
> That's true, though `checkpoint` does need to defuse it so it's also not as simple as defusing only once.

With the drop bomb, `checkpoint` wants to consume the `Accumulator`, and return `Result<Accumulator>`.  Otherwise if `checkpoint` returns `Err` the caller gets to keep the (necessarily defused) `Accumulator`, which ought to panic if it were used, and we want to avoid panics on unlikely error paths.

TedDriggs:
We're getting very close here.

I think we can do better than `run` and `handle` for the two method names. Maybe `take_err` for `handle` and `take_err_from` for `run`?

Part of my gripe with those two names is that we encourage this to be called `errors` when it's in scope. `errors.run(...)` doesn't inspire much confidence when I'm reading it.

ijackson:
To my mind `take_err` would *return* an `err`, not squirrel errors away and return the success value.  Cf `Option::take` (which extracts data from its receiver - almost the opposite of `handle`); `TcpStream::take_error` (which returns an `Error`), `take_pidfd`, etc.

I see your problem with `run`.  I think there's no problem with `errors.handle(something)`.

How about renaming `run` to `handle_from` ?

```rust
    errors.handle_from(||{
        let validated = thing.validate()?;
        outputs.push(validated);
        Ok(())
    });
```

`noun.transitive_verb(wombat)` often means "ask `noun` to `transitive_verb` this `wombat`".  `handle` fits that well.  "Ask the `errors` to handle this result".

TedDriggs:
I'm officially in love with the use of `wombat` for placeholder variable names.

---

You're right about `take` - I was thinking about its effect on the residual, e.g. "I took the error, so now what's left is the `Ok` value or a `None`" but that isn't quite what the std lib ones do.

---

I think I prefer `handle_in` or `handle_with` over `handle_from`, mostly because of the strong conversion connotation that `from` has in Rust. Then it's `errors.handle_in(|| { ... })`.

Is it bad that I actually quite like `errors.deal_with<T>(v: Result<T>) -> Option<T>`? It's probably too colloquial, but it does read quite clearly in American English. The closure version would be `errors.deal_with_in` or `errors.deal_with_from`.

ijackson:
"Deal with" is fine in British English too.  It is very slightly less formal, I think.  My only concern with it is whether non-native speakers of English would find it confusing (especially because of the way Rust uses `with`).

I looked at Wiktionary's entries for [deal with](https://en.wiktionary.org/wiki/deal_with) and [handle](https://en.wiktionary.org/wiki/handle#Etymology_2).  "Deal with" less ambiguous if you know what it means at all (since it can't mean "get *a handle*", but as I say I wonder if it's too confusing if you don't.  I did a search of some random local rustdocs here "handle" as in "deal with" is used a fair bit in web frameworks as a thing you do to requests or, indeed, errors.

Of the things listed in Wiktionary's synonyms sections for these words (I tried an online thesaurus but it was no use) I think "manage" is a possibility.  Also maybe "consider"

TedDriggs:
I'm supportive of `handle` and `handle_in`.

core/src/error/mod.rs
```diff
@@ -491,6 +500,177 @@ impl Iterator for IntoIter {
...
+            Ok(success)
+        } else {
+            Err(Error::multiple(errors))
+        }
+    }
+
+    fn errors(&mut self) -> &mut Vec<Error> {
+        match &mut self.0 {
+            Some(errors) => errors,
+            None => panic!("darling internal error"),
```
TedDriggs, line 581:
```suggestion
            None => panic!("darling internal error: Accumulator accessed after defuse"),
```

core/src/error/mod.rs
```diff
@@ -491,6 +500,177 @@ impl Iterator for IntoIter {
...
+    }
+
+    /// Returns the accumulated errors as a `Vec`.
+    ///
+    /// This function defuses the drop bomb.
+    #[must_use = "Accumulated errors should be handled or propagated to the caller"]
+    pub fn into_inner(mut self) -> Vec<Error> {
+        match std::mem::replace(&mut self.0, None) {
+            Some(errors) => errors,
+            None => panic!("darling internal error"),
```
TedDriggs, line 592:
```suggestion
            None => panic!("darling internal error: Accumulator accessed after defuse"),
```

TedDriggs (approved):
This looks great; really appreciate all the work you've put into it, and think this is a great addition to `darling`.

ijackson:
> This looks great; really appreciate all the work you've put into it, and think this is a great addition to `darling`.

Thanks for your kind words :-).  You're very welcome.

PR #168, opened 2022-03-27
fix: small error of consume_fields.rs in examples

BinChengZhao:
Hi, 

I was learning `darling` and found a small code error in consume_fields.rs in examples.

The `write!` macro needs to pass the corresponding Formatter instance writer as an argument.

PS: I found this when I implemented similar code locally~


TedDriggs:
Thanks for the PR! Writing examples whose output is itself tested is something I haven't yet figured out how to do, so checks like this are much appreciated.

BinChengZhao:
Enjoyed learning the `darling` process.

: )

PR #178, opened 2022-04-12
Handle validation of empty enum correctly

sreenathkrishnan:


TedDriggs:
Thanks for the submission! I think I'm understanding that the bug here is "shape validation will incorrectly allow enums that have no variants when only `struct_*` shapes are declared supported" - is that right?

sreenathkrishnan:
Yep, that's right!

core/src/options/shape.rs
```diff
@@ -72,13 +72,25 @@ impl ToTokens for Shape {
...
                 match *__body {
                     ::syn::Data::Enum(ref data) => {
                         fn validate_variant(data: &::syn::Fields) -> ::darling::Result<()> {
                             #en
                         }
 
+                        if data.variants.is_empty() {
+                            #check_empty_enum
+                        }
+
```
TedDriggs, line 93: If the receiver doesn't support any enum shapes, should we surface one diagnostic rather than one per variant?

sreenathkrishnan: Yeah, would be better to surface one diagnostic.

sreenathkrishnan: Updated!

TedDriggs (approved):

PR #186, opened 2022-05-03
Add ast::Meta<T>

TedDriggs:
This is generic over the literal expected by syn::Meta, allowing
attributes that use the meta structure but allow arbitrary syntax as
rvalues to be parsed by darling with minimal custom code.


TedDriggs:
This is no longer necessary with syn 2

PR #193, opened 2022-07-24
Error when parsing FromMeta containing NestedMeta::Lit

### LINKED ISSUE #192: FromMeta parsing doesn't error on unrecognized literals in attribute list
FromMeta doesn't error in all cases that unexpected tokens end up in the attribute list.

It doesn't error in the case that a string literal is in the attribute list. For example:
```rust
#[myattribute(myfield = "export", "randomliteral")]
```

Parsing the attribute list will error if it sees a field that is anticipated, but not for `"randomliteral"` which it will ignore.

cc @jonjove

leighmcculloch:
### What
Generate an error when parsing derived `FromMeta` implementations that contain a `NestedMeta::Lit`.

### Why
Literal values are unsupported in the code generated by `FromMeta`. Literal values, like strings, get ignored by the parse implementation that gets generated with the FromMeta generated code. This is confusing to a user since it appears like the value is being accepted and used in some way. We should error when the user provides a literal since the derived code will do nothing with it.

Fix #192

### DIFF
--- a/core/src/codegen/variant_data.rs
+++ b/core/src/codegen/variant_data.rs
@@ -64,6 +64,8 @@ impl<'a> FieldsGen<'a> {
                         #(#arms)*
                         __other => { #handle_unknown }
                     }
+                } else {
+                    __errors.push(::darling::Error::unsupported_format("literal").with_span(__item));
                 }
             }
         )
--- a/core/src/codegen/variant_data.rs
+++ b/core/src/codegen/variant_data.rs
@@ -66,8 +66,9 @@ impl<'a> FieldsGen<'a> {
                             __other => { #handle_unknown }
                         }
                     }
-                    ::syn::NestedMeta::Lit(_) => {
-                        __errors.push(::darling::Error::unsupported_format("literal").with_span(__item));
+                    ::syn::NestedMeta::Lit(ref __inner) => {
+                        __errors.push(::darling::Error::unsupported_format("literal")
+                            .with_span(__inner));
                     }
                 }
             }
--- a/core/src/codegen/variant_data.rs
+++ b/core/src/codegen/variant_data.rs
@@ -58,14 +58,17 @@ impl<'a> FieldsGen<'a> {
 
         quote!(
             for __item in __items {
-                if let ::syn::NestedMeta::Meta(ref __inner) = *__item {
-                    let __name = ::darling::util::path_to_string(__inner.path());
-                    match __name.as_str() {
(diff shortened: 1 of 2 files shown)


leighmcculloch:
👋🏻

TedDriggs:
Thanks for the reminder! I’m away from a computer for a few days here, but will finish reviewing when I am back. Apologies for the slow turnaround.

core/src/codegen/variant_data.rs
```diff
@@ -58,11 +58,17 @@ impl<'a> FieldsGen<'a> {
...
+                    ::syn::NestedMeta::Meta(ref __inner) => {
+                        let __name = ::darling::util::path_to_string(__inner.path());
+                        match __name.as_str() {
+                            #(#arms)*
+                            __other => { #handle_unknown }
+                        }
+                    }
+                    ::syn::NestedMeta::Lit(ref __inner) => {
+                        __errors.push(::darling::Error::unsupported_format("literal")
+                            .with_span(__inner));
```
TedDriggs, line 71: Should this honor `allow_unknown_fields` and avoid an error? I'm a bit worried that this change might break a use-case where `darling` does a partial parse, and thinking that the existing flag for this would cover that scenario. 

If there are reasons we shouldn't do that, we _could_ add a new `allow_literals`, but I'd want to see more evidence of a need for another flag before we expand the API surface.

TedDriggs (approved):
One small point of discussion, otherwise looks great.

core/src/codegen/variant_data.rs
leighmcculloch: > I'm a bit worried that this change might break a use-case where `darling` does a partial parse

Would a partial parse result in a nested meta being interpreted as a literal rather than a meta? A literal would always come after a path so it's less clear to me how we'd get a literal without the path in a partial parse.

On the one hand respecting `allow_unknown_fields` here could make sense if in the future darling is going to support literals are params.

On the other hand it doesn't really make sense because today if you pass a literal to the macro you're using the macro incorrectly.

Happy to do what you think is best here.

TedDriggs: I think leaving at as-shown in the PR is reasonable.

leighmcculloch:
> I think leaving at as-shown in the PR is reasonable.

👍🏻 

Anything I need to do for this PR for it to be mergeable?

leighmcculloch:
Thanks for merging!

Issue #197, opened 2022-09-05
Simple examples


casey:
I've had a hard time figuring out how to use this crate, both by reading the documentation and looking at the examples. It would be nice if there were simple examples that showed one feature at a time, or a guide with similarly simple example code.

PR #199, opened 2022-09-11
Add example to docs

casey:
Here's an example of the type of example that would have helped me figure out how to use darling, see #197 

- Shows the derive macro input
- Shows the darling parser
- Extracts a single top-level attribute

A sequence of examples such as these, each one performing a single task, and exercising a single feature of darling, would make it easy to get started.

I think that doctests are the ideal place to put these, for discoverability and ease of browsing, but there are multiple difficulties:

- Doctests cannot use #[proc_macro_derive]
- Doctests can't contain both the derive invocation and parser

But I wanted to open this PR as a starting point for discussion.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,9 @@
 //! # Darling
+//!
 //! Darling is a tool for declarative attribute parsing in proc macro implementations.
 //!
-//!
 //! ## Design
+//!
 //! Darling takes considerable design inspiration from [`serde`](https://serde.rs). A data structure that can be
 //! read from any attribute implements `FromMeta` (or has an implementation automatically
 //! generated using `derive`). Any crate can provide `FromMeta` implementations, even one not
@@ -12,6 +13,7 @@
 //! `FromField`, `FromVariant`, `FromGenerics`, _et alia_ to gather settings relevant to their operation.
 //!
 //! ## Attributes
+//!
 //! There are a number of attributes that `darling` exposes to enable finer-grained control over the code
 //! it generates.
 //!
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -63,6 +65,40 @@
...
+//! #[derive(FromDeriveInput)]
+//! struct Foo {
+//!   bar: syn::LitStr,
+//! }
+//!
+//! #[proc_macro_derive(Foo, attributes(foo))]
+//! pub fn display(item: proc_macro::TokenStream) -> proc_macro::TokenStream {
+//!   let derive_input = syn::parse_macro_input!(item as syn::DeriveInput);
+//!
+//!   let foo = Foo::from_derive_input(&derive_input).unwrap();
```
TedDriggs, line 97: We shouldn't use `unwrap` in an example. There's a right way to check the errors which will cause them to be emitted as compile errors; I have to look for where it's demonstrated, since it's been a while since I looked at this piece of the code.

src/lib.rs
```diff
@@ -1,8 +1,9 @@
 //! # Darling
-//! Darling is a tool for declarative attribute parsing in proc macro implementations.
 //!
+//! Darling is a tool for declarative attribute parsing in proc macro implementations.
```
TedDriggs, line 3: Are these whitespace changes necessary?

TedDriggs (changes requested):
Thanks for improving this! I've left a few comments; the error handling is the biggest one; I'm trying to find a great example of the right way to do it.

src/lib.rs
TedDriggs: [Here is a better approach from `derive_builder_core`](https://github.com/colin-kiegel/rust-derive-builder/blob/master/derive_builder_core/src/lib.rs#L62-L67)

```rust
    let opts = match macro_options::Options::from_derive_input(&ast) {
        Ok(val) => val,
        Err(err) => {
            return err.write_errors();
        }
    };
```

The `write_errors()` call produces a `proc_macro2::TokenStream`, so calling `into` on it will then convert it into the `proc_macro::TokenStream` expected by the derive entry-point function.

src/lib.rs
casey: No, but they do help make the doc comment more readable in an editor. I can remove them if you'd prefer.

src/lib.rs
```diff
@@ -63,6 +65,40 @@
...
+//!
+//! ### Derive Macro
+//!
+//! Here's an example-invocation of a derive macro `Foo` that parses a single top-level attribute
+//! named `bar` that accepts a string value:
+//!
+//! ```ignore
+//! #[derive(Foo)]
+//! #[foo(bar = "hello")]
+//! struct Foo {
```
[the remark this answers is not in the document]
casey: I actually think `Foo`, `foo`, and `bar` are useful here. Darling assigns special meaning to a few names, and this makes it clear that these are names which are up to the user to decide. I did go ahead and rename `struct Foo` to `struct Input`.

casey:
Added some commits addressing your comments. I didn't remove the whitespace changes, they're just to improve readability, but I can undo those or put them in another PR.

casey:
@TedDriggs Friendly ping!

TedDriggs:
Apologies for the delay - I'll be busy with work for the next few days, but haven't forgotten about this and will get back to it as soon as I can.

PR #206, opened 2022-11-23
Don't directly reference ::syn in generated code

### LINKED ISSUE #202: Expects the syn crate to be imported by the user as the name syn
While creating the reproduction for #201, I originally did not add `syn` as my own dependency:

```toml
[package]
name = "repro"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
darling = "0.14.1"
```

This causes the error:

```
error[E0433]: failed to resolve: could not find `syn` in the list of imported crates
 --> src/main.rs:1:10
  |
1 | #[derive(darling::FromVariant)]
  |          ^^^^^^^^^^^^^^^^^^^^ could not find `syn` in the list of imported crates
  |
  = note: this error originates in the derive macro `darling::FromVariant` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Ideally, darling would not expect that the user has a crate named `syn` in scope. A better idea is to re-export syn from the darling crate and then you can use that path without worry.

A related problem is if darling _itself_ is renamed. SNAFU uses the [`crate_root`](https://docs.rs/snafu/latest/snafu/derive.Snafu.html#controlling-how-the-snafu-crate-is-resolved) attribute to allow the end user to specify what the `snafu` crate should actually be called.

TedDriggs:
Fixes #202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Re-export `syn` from `darling` to avoid requiring that consuming crates have a `syn` dependency.
+
 ## v0.14.2 (October 26, 2022)
 
 - Derived impls of `FromMeta` will now error on literals, rather than silently ignoring them. [#193](https://github.com/TedDriggs/darling/pull/193)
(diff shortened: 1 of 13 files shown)


TedDriggs:
@shepmaster this fixes the narrower issue described in #202; I think a separate issue for allowing rename/re-export of `darling` is warranted because it's a much bigger code change.

PR #211, opened 2022-11-30
Keep span information for use of Default::default

TedDriggs:
This improves error messages when generated code tries to get a default value and one isn't supported.

### DIFF
--- a/core/src/options/core.rs
+++ b/core/src/options/core.rs
@@ -66,9 +66,13 @@ impl Core {
     fn as_codegen_default(&self) -> Option<codegen::DefaultExpression<'_>> {
         self.default.as_ref().map(|expr| match *expr {
             DefaultExpression::Explicit(ref path) => codegen::DefaultExpression::Explicit(path),
-            DefaultExpression::Inherit | DefaultExpression::Trait => {
-                codegen::DefaultExpression::Trait
+            DefaultExpression::Inherit => {
+                // It should be impossible for any input to get here,
+                // so panic rather than returning an error or pretending
+                // everything is fine.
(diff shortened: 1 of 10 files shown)


core/src/options/core.rs
```diff
@@ -66,9 +66,13 @@ impl Core {
...
         self.default.as_ref().map(|expr| match *expr {
             DefaultExpression::Explicit(ref path) => codegen::DefaultExpression::Explicit(path),
-            DefaultExpression::Inherit | DefaultExpression::Trait => {
-                codegen::DefaultExpression::Trait
+            DefaultExpression::Inherit => {
+                // It should be impossible for any input to get here,
+                // so panic rather than returning an error or pretending
+                // everything is fine.
+                panic!("DefaultExpression::Inherit is not valid at container level")
             }
```
TedDriggs, line 74: Outside the scope of this PR, it feels like it should be possible to rework this and remove any chance of macro-runtime failure here.

core/src/options/input_field.rs
```diff
@@ -97,11 +98,7 @@ impl InputField {
         // 1. Will we look for this field in the attribute?
         // 1. Is there a locally-defined default?
         // 1. Did the parent define a default?
-        self.default = match (
-            self.skip.unwrap_or_default(),
-            self.default.is_some(),
-            parent.default.is_some(),
-        ) {
+        self.default = match (&self.skip, self.default.is_some(), parent.default.is_some()) {
```
TedDriggs, line 101: An NFC to clean this up seems sensible:

- Make the first case be a simple if-statement, since it overrules the others
- Possibly rework second and third variables as well?

PR #220, opened 2023-01-10
Define FromData trait

smmoosavi:
define `FromData` trait, so the user can defines custom type for the data field in `FromDeriveInput`

example:

```rust

#[derive(Debug, FromDeriveInput)]
#[darling(attributes(from_variants))]
pub struct Container {
    pub data: MyData,
}

#[derive(Default, Debug, FromVariant)]
#[darling(default, attributes(from_variants))]
pub struct Variant {
    into: Option<bool>,
    skip: Option<bool>,
}

#[derive(Debug)]
pub struct MyData {
    pub variants: Vec<Variant>,
}

impl FromData for MyData {
    fn from_data(data: &syn::Data) -> darling::Result<Self> {
        let data: ast::Data<Variant, ()> = FromData::from_data(data)?;
        match data {
            ast::Data::Enum(variants) => Ok(MyData { variants }),
            ast::Data::Struct(_) => Err(darling_core::Error::unsupported_shape("struct")),
        }
    }
}
```

### DIFF
--- a/core/src/ast/data.rs
+++ b/core/src/ast/data.rs
@@ -4,6 +4,7 @@ use proc_macro2::{Span, TokenStream};
 use quote::{quote, quote_spanned, ToTokens};
 use syn::spanned::Spanned;
 
+use crate::from_data::FromData;
 use crate::usage::{
     self, IdentRefSet, IdentSet, LifetimeRefSet, LifetimeSet, UsesLifetimes, UsesTypeParams,
 };
@@ -136,6 +137,12 @@ impl<V: FromVariant, F: FromField> Data<V, F> {
             // putting it on the union keyword ends up being confusing.
             syn::Data::Union(_) => Err(Error::custom("Unions are not supported")),
         }
+    }
+}
+
+impl<V: FromVariant, F: FromField> FromData for Data<V, F> {
+    fn from_data(data: &syn::Data) -> Result<Self> {
+        Self::try_from(data)
     }
 }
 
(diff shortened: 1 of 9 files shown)


smmoosavi:
I missed implementing `FromData` for ignored. I will add It in a few hours.

TedDriggs:
What problem does this solve that isn't addressed by `darling::ast::Data` and shape validation on the `FromDeriveInput`-deriving struct?

smmoosavi:
shape validation only shows errors but types are the same. but when we want to use the Data we should use `.take_enum().unwrap()` or `.take_struct().ok_or_else(...)` to handle it. with the `FromData`, we have an easy and safe code.

smmoosavi:
the `#[darling(map="...")]` can be used for data or other base fields?

TedDriggs:
> shape validation only shows errors but types are the same. but when we want to use the Data we should use .take_enum().unwrap() or .take_struct().ok_or_else(...) to handle it. with the FromData, we have an easy and safe code.

This could be addressed with the `!` type once that's stabilized:

```rust
impl<F> Data<!, F> {
    fn fields(self) -> Fields<F> {}
}

impl<V> Data<V, !> {
    fn variants(self) -> Vec<V> {}
}
```

As it stands right now, `darling` supports adaptation of `syn` types _either_ through derive-macros and magic fields _or_ through generic equivalents of `syn` types. This would be the first place where we offer both support models, and I'm worried that would be confusing to people.

TedDriggs:
With let-else being a thing now, it’s easier than ever to use `ast::Data` without causing increased indentation or other issues. I’d recommend that instead of adding another derive-macro.

PR #223, opened 2023-02-01
Expose expected shapes in shape validation

### LINKED ISSUE #222: Include supported shapes in unsupported shape error messages
I cannot match the kind of error because it is `not public` to external crate.

https://github.com/TedDriggs/darling/blob/bafada6f97fd67dfba9005e47934f80b66ec7977/core/src/error/kind.rs#L12-L13

I wanted to give custom error message when `king = UnsupportedShape`.

TedDriggs:
Fixes #222

### DIFF
--- a/core/src/options/shape.rs
+++ b/core/src/options/shape.rs
@@ -3,7 +3,7 @@
 
 use proc_macro2::TokenStream;
 use quote::{quote, ToTokens, TokenStreamExt};
-use syn::{Meta, NestedMeta};
+use syn::{parse_quote, Meta, NestedMeta};
 
 use crate::{Error, FromMeta, Result};
 
@@ -11,24 +11,19 @@ use crate::{Error, FromMeta, Result};
 /// to declare that it only accepts - for example - named structs, or newtype enum
 /// variants.
 ///
-/// # Usage
-/// Because `Shape` implements `FromMeta`, the name of the field where it appears is
-/// controlled by the struct that declares `Shape` as a member. That field name is
-/// shown as `ignore` below.
-///
 /// ```rust,ignore
 /// #[ignore(any, struct_named, enum_newtype)]
 /// ```
 #[derive(Debug, Clone)]
-pub struct Shape {
+pub struct DeriveInputShapeSet {
     enum_values: DataShape,
     struct_values: DataShape,
     any: bool,
 }
 
-impl Default for Shape {
+impl Default for DeriveInputShapeSet {
     fn default() -> Self {
-        Shape {
+        DeriveInputShapeSet {
             enum_values: DataShape::new("enum_"),
             struct_values: DataShape::new("struct_"),
             any: Default::default(),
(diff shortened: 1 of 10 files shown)


core/src/util/shape.rs
```diff
@@ -0,0 +1,243 @@
...
+    tuple: bool,
+    unit: bool,
+    any: bool,
+}
+
+impl ShapeSet {
+    pub fn set_any(&mut self, any: bool) -> &mut Self {
+        self.any = any;
+        self
+    }
```
TedDriggs, line 102: This isn't being used by the generated code - is it necessary?

core/src/util/shape.rs
```diff
@@ -0,0 +1,243 @@
...
+            Err(crate::Error::unsupported_shape_with_expected(
+                shape.description(),
+                self,
+            ))
+        }
+    }
+}
+
+impl fmt::Display for ShapeSet {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
TedDriggs, line 151: It would be nice to build this as a more fluent sentence, e.g. "named fields or unnamed fields" in the case of two items, or "named fields, unnamed fields, or no fields" when three options are applicable.

core/src/util/shape.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+impl AsShape for Shape {
+    fn as_shape(&self) -> Shape {
+        *self
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct ShapeSet {
```
TedDriggs, line 90: This needs a doc comment.

core/src/options/shape.rs
```diff
@@ -12,33 +12,33 @@ use crate::{Error, FromMeta, Result};
...
+/// Because `ShapeOpt` implements `FromMeta`, the name of the field where it appears is
+/// controlled by the struct that declares `ShapeOpt` as a member. That field name is
 /// shown as `ignore` below.
 ///
 /// ```rust,ignore
 /// #[ignore(any, struct_named, enum_newtype)]
 /// ```
 #[derive(Debug, Clone)]
-pub struct Shape {
+pub struct ShapeOpt {
```
TedDriggs, line 23: This is not a good name.

core/src/util/shape.rs
TedDriggs, line 90: Should this provide its own `impl FromMeta` for convenience if people want to use it in their crates?

TedDriggs:
@Diwakar-Gupta, want to take a look at this before it merges?

core/src/options/shape.rs
Diwakar-Gupta: I'm not sure, but altering the name might make earlier versions incompatible.?

core/src/util/shape.rs
TedDriggs: It's not; the `any` field has been removed in favor of adding the `ShapeSet::insert_all` method.

core/src/util/shape.rs
TedDriggs: Done

core/src/util/shape.rs
TedDriggs: Done.

core/src/options/shape.rs
TedDriggs: This struct isn't visible outside `darling_core`, so that isn't a risk. The type is a bit weird because it's basically smashing two macro-time `ShapeSet` values together, along with the `any` field. I decided to go with a name based on how it's used, and called it `DeriveInputShapeSet`.

core/src/util/shape.rs
TedDriggs: Decided that doing this is not a must-have for the initial release, and adding trait impls now is a breaking change to undo later.

PR #225, opened 2023-03-08
Add child diagnostics

### LINKED ISSUE #224: Support for custom diagnostics in Error
Hello! Great work on this crate.

First of all, I am aware of #175 being opened. I just don't know if this is the same issue, as the other one mentions non-fatal diagnostics, so I'm creating a new one. Feel free to close this if non-applicable.

I can see that there's support for diagnostics through a feature. Namely, `did_you_mean` will show alternatives as a help diagnostic.

I've been writing proc macros at work for a custom use case, and wanted to refactor said macro to use `darling` as we believe it brings value (easier maintainance, better readability). The macro currently has custom diagnostics, implemented through a custom error enum +  a match block that leverags `proc_macro_error` to display error diagnostics. This is important to us, because it immediately clarifies what's wrong when working with our own macros.

One issue that's preventing us from refactoring with darling is the lack of custom diagnostics - I'm speaking about the possibility to attach a custom diagnostic to a `darling` Error like `proc_macro_error` does. The API to attach custom diagnostics could for example look like this : 

`Error::custom("My custom error").with_diagnostic("note", "My note on the macro usage").with_diagnostic("help", "Try doing this instead")`

Which would then display the following error at compile-time : 

```
error: My custom error
  --> my_project/my_file.rs:3:5
   |
13 |     FooBar { value: String },
   |     ^^^^^^
   |
   = note: My note on the macro usage
   = help: Try doing this instead
```

In terms of implementation, this would probably require adding a Vec of custom diagnostics at the Error level. I dont know how feasible this is, but I could have a go at implementing it if there's interest ?

TedDriggs:
Fixes #224

# Open Items
- [x] Update release notes
- [x] Add unit tests ([details](https://github.com/TedDriggs/darling/pull/225#issuecomment-1462046052))
- [x] Update CI to run diagnostics unit tests

### DIFF
--- a/core/src/error/child.rs
+++ b/core/src/error/child.rs
@@ -1,0 +1,82 @@
+use proc_macro2::Span;
+
+/// Exhaustive mirror of [`proc_macro::Level`].
+#[derive(Debug, Clone)]
+pub(in crate::error) enum Level {
+    Error,
+    Warning,
+    Note,
+    Help,
+}
+
+/// Supplemental message for an [`Error`](super::Error) when it's emitted as a `Diagnostic`.
+///
+/// # Example Output
+/// The `note` and `help` lines below come from child diagnostics.
+///
+/// ```text
+/// error: My custom error
+///   --> my_project/my_file.rs:3:5
+///    |
+/// 13 |     FooBar { value: String },
+///    |     ^^^^^^
+///    |
+///    = note: My note on the macro usage
+///    = help: Try doing this instead
+/// ```
+#[derive(Debug, Clone)]
+pub(in crate::error) struct ChildDiagnostic {
+    level: Level,
+    span: Option<Span>,
+    message: String,
+}
+
+impl ChildDiagnostic {
+    pub(in crate::error) fn new(level: Level, span: Option<Span>, message: String) -> Self {
+        Self {
+            level,
+            span,
+            message,
+        }
+    }
+}
+
+impl ChildDiagnostic {
+    /// Append this child diagnostic to a `Diagnostic`.
+    ///
+    /// # Panics
+    /// This method panics if `self` has a span and is being invoked outside of
+    /// a proc-macro due to the behavior of [`Span::unwrap()`](Span).
+    pub fn append_to(self, diagnostic: proc_macro::Diagnostic) -> proc_macro::Diagnostic {
+        match self.level {
+            Level::Error => {
+                if let Some(span) = self.span {
+                    diagnostic.span_error(span.unwrap(), self.message)
+                } else {
+                    diagnostic.error(self.message)
+                }
+            }
+            Level::Warning => {
+                if let Some(span) = self.span {
+                    diagnostic.span_warning(span.unwrap(), self.message)
+                } else {
+                    diagnostic.warning(self.message)
+                }
+            }
+            Level::Note => {
+                if let Some(span) = self.span {
+                    diagnostic.span_note(span.unwrap(), self.message)
+                } else {
+                    diagnostic.note(self.message)
+                }
+            }
+            Level::Help => {
+                if let Some(span) = self.span {
+                    diagnostic.span_help(span.unwrap(), self.message)
+                } else {
+                    diagnostic.help(self.message)
+                }
+            }
+        }
+    }
+}
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -15,6 +15,8 @@ use std::vec;
 use syn::spanned::Spanned;
 use syn::{Lit, LitStr, Path};
 
+#[cfg(feature = "diagnostics")]
+mod child;
 mod kind;
 
 use crate::util::path_to_string;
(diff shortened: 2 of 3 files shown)


core/src/error/mod.rs
```diff
@@ -234,6 +241,82 @@ impl Error {
     }
 }
 
+/// Child diagnostic methods
+#[cfg(feature = "diagnostics")]
+impl Error {
+    pub fn error<T: fmt::Display>(mut self, message: &T) -> Self {
```
AlisCode, line 247: NIT: all this code is copy-pasted and could be macro-generated for the sake of maintainablity. By the way, [it is what they do in the `proc_macro` library](https://github.com/rust-lang/rust/blob/master/library/proc_macro/src/diagnostic.rs#L57). I don't know if it's worth the trouble though :)

core/src/error/mod.rs
```diff
@@ -371,13 +472,17 @@ impl Error {
         //
         // If span information is available, don't include the error property path
         // since it's redundant and not consistent with native compiler diagnostics.
-        match self.kind {
+        let diagnostic = match self.kind {
             ErrorKind::UnknownField(euf) => euf.into_diagnostic(self.span),
             _ => match self.span {
```
AlisCode, line 477: I'd advise against using a discard match-arm here, because in case we end up adding a new variant that contains one or many diagnostics, they will get erased possibly without us noticing. Remove the discard here and the compiler will warn you instead.

AlisCode: btw I don't understand the rationale behind `ErrorKind` being "hacked to be non-exhaustive" (as in, not using stdlib's `#[non_exhaustive]` attribute) ? And is it a good idea to be non-exhaustive, even ?

I think it's fair for `darling` to say "this is the set of errors we currently support", what am I missing ?

AlisCode (commented):
This seems to do the work, good job! I've put some feedback, feel free to take it or not :)

core/src/error/child.rs
```diff
@@ -0,0 +1,75 @@
+use proc_macro::Level;
+use proc_macro2::Span;
+
+/// Supplemental message for an [`Error`](super::Error) when it's emitted as a `Diagnostic`.
+///
+/// # Example Output
+/// The `note` and `help` lines below come from child diagnostics.
```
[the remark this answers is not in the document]
TedDriggs: I think it makes sense to put examples on the methods themselves, but not on the crate-internal struct.

core/src/error/mod.rs
TedDriggs: This isn't a discard arm; we have special behavior for converting `ErrorUnknownField` into a `Diagnostic`, while for all the other `ErrorKind` variants we only need to use the `Display` impl of `ErrorKind` to get the diagnostic body.

> btw I don't understand the rationale behind ErrorKind being "hacked to be non-exhaustive" (as in, not using stdlib's #[non_exhaustive] attribute) ?

`darling` has an MSRV of 1.31; the `#[non_exhaustive]` attribute was [introduced in 1.40](https://blog.rust-lang.org/2019/12/19/Rust-1.40.0.html), so `darling` cannot use it.

> And is it a good idea to be non-exhaustive, even ?

It ends up not mattering at all, because `ErrorKind` is always going to remain internal to the `error` mod. The comments make it look like I was considering having `ErrorKind` be public at some point, but that was at least 4 years ago and I'm now 100% confident that `ErrorKind` will always stay internal. If you're interested in submitting a PR that removes the `__NonExhaustive` variant, I'd be happy to review one.

> I think it's fair for darling to say "this is the set of errors we currently support", what am I missing ?

It matters a lot that `darling` be able to add new error variants or _change the internal representation of error variants_ without any risk of breaking crate consumers or needing a minor version bump:
* Features like the "did you mean" suggestions would be much harder to implement if I had to worry about breaking code that tried to match on specific error types that `darling` supports.
* Adoption of features like "did you mean" would be slower if that was pushed as a minor version bump instead of a patch version bump.
* I think it's quite valuable that the output from using `Error::custom` is indistinguishable from the output of using other `Error` constructor functions, and that wouldn't be possible if people could interact with `ErrorKind`.

core/src/error/mod.rs
TedDriggs: Fixed.

TedDriggs:
@AlisCode Rather than always pinning nightly, I'd suggest adding a specific new job that tests diagnostics with that feature enabled and the nightly version pinned. That'll avoid changing the current test slate which runs against the latest nightly.

I think the best solution for the panic issue is to manually test `append_to` now and then just hope for the best going forward 😅 The unit tests can focus on making sure the ergonomics of adding these messages to an error work well.

> Thanks for the reactivity and the good work !

Thanks for the great feature request!

core/src/error/child.rs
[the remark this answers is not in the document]
TedDriggs: I've added examples to the relevant impl block for `Error` and to each of the public methods.

core/src/error/mod.rs
TedDriggs: I made this into a macro. The code is now much nicer to read - thanks for the suggestion.

TedDriggs:
@AlisCode the doc tests I added seem to do a good enough job of making sure the happy-path compiles, so I'm going to declare this "good enough for a feature-flagged feature" and merge.

Issue #229, opened 2023-04-27
Allow string literals for meta items of type syn::Expr


TedDriggs:
Right now, when `darling` encounters a string literal for a field of type `syn::Expr`, it will attempt to parse the string contents and put that result into the struct. This means it's impossible to have a field that takes an arbitrary expression that may happen to be a string literal.

# Proposed Fix
None yet; please comment on this issue if you're relying on the current behavior or want to rely on the behavior described above.

milesj:
Been thinking about this a bit more. What if there was another derive, like `FromAttributesStrict`, that preserved string literals. This would be backwards compatible.

For context, I'm using `Expr` to define default values using a `default` attr meta: https://github.com/moonrepo/schematic/blob/master/crates/config/tests/macros_test.rs#L46 Because of this string problem, I have to add 2 additional meta fields, `default_str` and `default_fn`.

TedDriggs:
I'm leaning towards a solution that uses `#[darling(with="...")]` to choose a parsing function with the desired handling of string literals.

I don't want to forbid quotation marks in places where they don't cause ambiguity; if nothing else, as long as serde has quotation marks around its paths I'm going to keep typing them by reflex, and breaking that unnecessarily would annoy maintainers and users of darling-powered macros. That means the `impl FromMeta for syn::Path` would continue to be indifferent about quotation marks. 

I think that syn::Expr is unique in having this problem: syn::Lit should be fine since it was previously valid in this position in syn 1, so we didn't have this issue of needing to "smuggle" content with quotation marks.

@milesj how about we aim to introduce a helper function in 0.20.1 that you can use to get unstuck, and then continue exploring the problem space to find the best long-term default?

milesj:
I'm cool with whatever approach allows me to have a single `default` field.

I've gone back and forth on whether to use strings for paths (like serde), or use actual paths, and landed on paths because it has better editor integration. For example, you can hover to see the signature, and click through to go to source. I wish more macros would follow suit.

<img width="666" alt="Screenshot 2023-05-18 at 11 03 42 AM" src="https://github.com/TedDriggs/darling/assets/143744/6a9592b8-bb01-4ce9-b62d-806e376e7938">

milesj:
Just chiming in that it's been working great. Thanks for the quick turn around 👍

greyblake:
At the moment I am rewriting [nutype](https://github.com/greyblake/nutype) parsing logic to use `darling` and I stumble upon this in a few different places.
[...]

PR #235, opened 2023-05-22
Add new parsing util functions for Exprs

TedDriggs:
This contributes to #229 by allowing crate users to more concisely choose the desired parsing behavior.


TedDriggs:
@milesj pointed out [here](https://github.com/TedDriggs/darling/issues/229#issuecomment-1557558000) that these functions are not generic over wrapper types such as `Option`, which is unfortunate.

TedDriggs:
It looks like these can be combined with the `map` field, so this works:

```rust
#[darling(with = "darling::util::parse_expr::preserve_str_literal", map = "Some")]
```

PR #243, opened 2023-07-11
Implement FromMeta for i128 and u128

greyblake:
While rewriting nutype's parsing logic, I face the fact that `FromMeta` is missing for `i128` and `u128`.
If you merge this I'd also appreciate a lot a release of a new patch version.

Thank you!


TedDriggs:
Will `syn` handle u128/i128 literals?

TedDriggs:
Publishing of 0.20.3 will happen once CI completes.

Issue #63, opened 2019-02-12
Add #[darling(word)] for enum variants


TedDriggs:
`DefaultExpression` is one example of an enum that wants to map a specific variant to the appearance of the field ident with no associated value. This would only be valid on variants in `FromMeta`. The variant must be a unit variant.

TedDriggs:
@chipsenkbeil I think this is what you're looking for `ent(field)`

chipsenkbeil:
In my case, I'm looking for a word or a list as the attribute can be `#[ent(field)]` if just marking a struct field as an *entity field*, or something like `#[ent(field(indexed, mutable))]` if marking as an entity field with special flags.

Separately, I also have `#[ent(edge)]` that can have `#[ent(edge(type = "some::type", policy = "SomeEnumVariant"))]`.

To summarize, I'm both looking for the presence of `field` or `edge` as a boolean indicator as well as optional inner words/named values to assign other properties. All of the inner content comes across as optional, with words defaulting to false as booleans and types defaulting to None/a default value.

TedDriggs:
Check out `darling::util::Override`. The naming isn't the most obvious for your use-case, but I believe it does what you're looking for.

TedDriggs:
Given the existence of the above workaround (and the lack of movement for almost 3 years), I'm closing this as "Won't Fix."

PR #260, opened 2023-12-28
Improve `FromMeta` implementation for enums

davidsemakula:
Improves FromMeta implementation for enums

- Properly implement unit arms in `FromMeta::from_list` to provide a consistent API for heterogenous enums that include a mix of unit, newtype and struct variants (see included test for an example).
- Add #[darling(word)] for unit enum variants (See #63).
- Adds tests for behavior of `FromMeta` implementation for enums.

### DIFF
--- a/core/src/codegen/from_meta_impl.rs
+++ b/core/src/codegen/from_meta_impl.rs
@@ -94,6 +94,21 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                     |default_expr| quote!(::darling::export::Ok(#default_expr)),
                 );
 
+                let word_or_err = variants
+                    .iter()
+                    .find_map(|variant| {
+                        if variant.word {
+                            let ty_ident = variant.ty_ident;
+                            let variant_ident = variant.variant_ident;
+                            Some(quote!(::darling::export::Ok(#ty_ident::#variant_ident)))
+                        } else {
+                            None
+                        }
+                    })
+                    .unwrap_or(quote!(::darling::export::Err(
+                        ::darling::Error::unsupported_format("word")
+                    )));
+
                 quote!(
                     fn from_list(__outer: &[::darling::export::NestedMeta]) -> ::darling::Result<Self> {
                         // An enum must have exactly one value inside the parentheses if it's not a unit
@@ -120,6 +135,10 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                             #(#unit_arms)*
                             __other => ::darling::export::Err(::darling::Error::unknown_value(__other))
                         }
+                    }
+
+                    fn from_word() -> ::darling::Result<Self> {
+                        #word_or_err
                     }
                 )
             }
--- a/core/src/codegen/from_meta_impl.rs
+++ b/core/src/codegen/from_meta_impl.rs
@@ -89,14 +89,10 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                     }
                 };
 
-                let default = base
-                    .default
-                    .as_ref()
-                    .map(|default_expr| quote!(::darling::export::Ok(#default_expr)));
-                let unit_arm_default = default.as_ref().map(|default| quote!("" => #default,));
-                let list_default_or_err = default.unwrap_or(quote! {
-                    ::darling::export::Err(::darling::Error::too_few_items(1))
-                });
+                let default_or_err = base.default.as_ref().map_or(
+                    quote!(::darling::export::Err(::darling::Error::too_few_items(1))),
+                    |default_expr| quote!(::darling::export::Ok(#default_expr)),
+                );
 
                 quote!(
                     fn from_list(__outer: &[::darling::export::NestedMeta]) -> ::darling::Result<Self> {
@@ -104,7 +100,7 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                         // match arm.
                         // NOTE: The empty list case uses the annotated default value for the enum if any.
                         match __outer.len() {
-                            0 => #list_default_or_err,
+                            0 => #default_or_err,
                             1 => {
                                 if let ::darling::export::NestedMeta::Meta(ref __nested) = __outer[0] {
                                     match ::darling::util::path_to_string(__nested.path()).as_ref() {
@@ -122,7 +118,6 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                     fn from_string(lit: &str) -> ::darling::Result<Self> {
                         match lit {
                             #(#unit_arms)*
-                            #unit_arm_default
                             __other => ::darling::export::Err(::darling::Error::unknown_value(__other))
                         }
                     }
--- a/core/src/codegen/from_meta_impl.rs
+++ b/core/src/codegen/from_meta_impl.rs
@@ -89,10 +89,13 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                     }
                 };
 
-                let default_or_err = base.default.as_ref().map_or(
-                    quote!(::darling::export::Err(::darling::Error::too_few_items(1))),
-                    |default_expr| quote!(::darling::export::Ok(#default_expr)),
-                );
+                let default_or_err = base
+                    .default
+                    .as_ref()
+                    .map(|default_expr| quote!(::darling::export::Ok(#default_expr)))
+                    .unwrap_or_else(|| {
+                        quote!(::darling::export::Err(::darling::Error::too_few_items(1)))
+                    });
 
                 let word_or_err = variants
                     .iter()
@@ -105,9 +108,11 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                             None
                         }
                     })
-                    .unwrap_or(quote!(::darling::export::Err(
-                        ::darling::Error::unsupported_format("word")
-                    )));
+                    .unwrap_or_else(|| {
+                        quote!(::darling::export::Err(
+                            ::darling::Error::unsupported_format("word")
+                        ))
+                    });
 
                 quote!(
                     fn from_list(__outer: &[::darling::export::NestedMeta]) -> ::darling::Result<Self> {
--- a/core/src/codegen/from_meta_impl.rs
+++ b/core/src/codegen/from_meta_impl.rs
@@ -77,7 +77,6 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
             }
             Data::Enum(ref variants) => {
                 let unit_arms = variants.iter().map(Variant::as_unit_match_arm);
-                let struct_arms = variants.iter().map(Variant::as_data_match_arm);
 
                 let unknown_variant_err = if !variants.is_empty() {
                     let names = variants.iter().map(Variant::as_name);
@@ -90,16 +89,26 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                     }
                 };
 
+                let default = base
+                    .default
+                    .as_ref()
+                    .map(|default_expr| quote!(::darling::export::Ok(#default_expr)));
+                let unit_arm_default = default.as_ref().map(|default| quote!("" => #default,));
+                let list_default_or_err = default.unwrap_or(quote! {
+                    ::darling::export::Err(::darling::Error::too_few_items(1))
+                });
+
                 quote!(
                     fn from_list(__outer: &[::darling::export::NestedMeta]) -> ::darling::Result<Self> {
                         // An enum must have exactly one value inside the parentheses if it's not a unit
-                        // match arm
+                        // match arm.
+                        // NOTE: The empty list case uses the annotated default value for the enum if any.
                         match __outer.len() {
-                            0 => ::darling::export::Err(::darling::Error::too_few_items(1)),
+                            0 => #list_default_or_err,
                             1 => {
                                 if let ::darling::export::NestedMeta::Meta(ref __nested) = __outer[0] {
                                     match ::darling::util::path_to_string(__nested.path()).as_ref() {
-                                        #(#struct_arms)*
+                                        #(#variants)*
                                         __other => ::darling::export::Err(::darling::Error::#unknown_variant_err.with_span(__nested))
                                     }
                                 } else {
@@ -113,6 +122,7 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                     fn from_string(lit: &str) -> ::darling::Result<Self> {
                         match lit {
                             #(#unit_arms)*
+                            #unit_arm_default
                             __other => ::darling::export::Err(::darling::Error::unknown_value(__other))
                         }
                     }
--- a/core/src/codegen/from_meta_impl.rs
+++ b/core/src/codegen/from_meta_impl.rs
@@ -89,14 +89,6 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                     }
                 };
 
-                let default_or_err = base
-                    .default
-                    .as_ref()
-                    .map(|default_expr| quote!(::darling::export::Ok(#default_expr)))
-                    .unwrap_or_else(|| {
-                        quote!(::darling::export::Err(::darling::Error::too_few_items(1)))
-                    });
-
                 let word_or_err = variants
                     .iter()
                     .find_map(|variant| {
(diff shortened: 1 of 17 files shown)


tests/from_meta.rs
```diff
@@ -64,3 +64,55 @@ fn nested_meta_lit_bool_errors() {
...
+mod enum_impl {
+    use darling::{Error, FromMeta};
+    use syn::parse_quote;
+
+    /// A playback volume.
+    /// Deriving `Default` will cause the "default" variant to be set when the meta-item is empty.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Default, FromMeta)]
+    #[darling(default)]
+    enum Volume {
+        #[default]
```
TedDriggs, line 78: I think the question of how to handle just the word form is discussed in #63; given that this isn't truly a default, I'd prefer to use the `#[darling(word)]` syntax rather than a bare `default` here.

core/src/codegen/from_meta_impl.rs
```diff
@@ -113,6 +122,7 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                     fn from_string(lit: &str) -> ::darling::Result<Self> {
                         match lit {
                             #(#unit_arms)*
+                            #unit_arm_default
```
TedDriggs, line 125: This change feels orthogonal to the heterogeneous arms improvement; can we split them into two PRs?

TedDriggs (changes requested):

core/src/codegen/from_meta_impl.rs
davidsemakula: makes sense ... will do.

tests/from_meta.rs
```diff
@@ -64,3 +64,55 @@ fn nested_meta_lit_bool_errors() {
...
 }
+
+/// Tests behavior of FromMeta implementation for enums (including behavior of defaults).
+mod enum_impl {
+    use darling::{Error, FromMeta};
+    use syn::parse_quote;
+
+    /// A playback volume.
+    /// Deriving `Default` will cause the "default" variant to be set when the meta-item is empty.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Default, FromMeta)]
```
TedDriggs, line 75: It looks like deriving `Default` on the enum breaks 1.56, which is part of Rust's MSRV. I don't think adding a derived `Default` (as opposed to an `impl` alters this case meaningfully, so I'd suggest adding a new `tests/enums_default.rs` folder and then having that showcase the two relevant behaviors:

1. Using `#[darling(default)]` on the receiver field
2. Using `#[darling(word)]` on the variant that is used when only a word is present.

```rust
use darling::{FromDeriveInput, FromMeta};
use syn::parse_quote;

#[derive(Debug, FromMeta, PartialEq, Eq)]
enum Dolor {
    Sit,
    #[darling(word)]
    Amet,
}

impl Default for Dolor {
    fn default() -> Self {
        Dolor::Sit
    }
}

#[derive(FromDeriveInput)]
#[darling(attributes(hello))]
struct Receiver {
    #[darling(default)]
    example: Dolor,
}

#[test]
fn missing_meta() {
    let di = Receiver::from_derive_input(&parse_quote! {
        #[hello]
        struct Example;
    })
    .unwrap();

    assert_eq!(Dolor::Sit, di.example);
}

#[test]
fn empty_meta() {
    let di = Receiver::from_derive_input(&parse_quote! {
        #[hello(example)]
        struct Example;
    })
    .unwrap();

    assert_eq!(Dolor::Amet, di.example);
}
```

davidsemakula: Will switch to an `impl` for `Default`.

Also I get this error when trying the above code snippet:
```
error: Unknown field: `word`
 --> tests/enums_default.rs:7:15
  |
7 |     #[darling(word)]
  |               ^^^^
```

TedDriggs: Makes sense that you'd get an error, since `#[darling(word)]` isn't implemented - that's the alternative syntax to `#[default]` that I proposed in another thread in this PR.

davidsemakula: I guess it was a bit of confusion on both our parts. 
The `#[default]` syntax in my example is actually from <https://doc.rust-lang.org/std/default/trait.Default.html#enums> so that's not a custom attribute I was adding.

davidsemakula:
I'm working on implementing support for [`#[darling(word)]` syntax](https://github.com/TedDriggs/darling/pull/260#discussion_r1439591569)

core/src/options/input_variant.rs
```diff
@@ -95,6 +98,12 @@ impl ParseAttribute for InputVariant {
             }
 
             self.skip = FromMeta::from_meta(mi)?;
+        } else if path.is_ident("word") && self.data.is_unit() {
```
TedDriggs, line 101: I'd suggest moving the second half of this check down; we should enter this branch for `word` and have a dedicated error message if the variant has fields.

core/src/codegen/from_meta_impl.rs
```diff
@@ -90,16 +89,37 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
...
 
+                let default_or_err = base.default.as_ref().map_or(
+                    quote!(::darling::export::Err(::darling::Error::too_few_items(1))),
+                    |default_expr| quote!(::darling::export::Ok(#default_expr)),
+                );
+
+                let word_or_err = variants
+                    .iter()
+                    .find_map(|variant| {
+                        if variant.word {
```
TedDriggs, line 100: There should be something that errors if two variants on the same enum both have `word` set; one of those is going to be ignored every time, and it's indicative of a dev error that darling should protect against.

Unfortunately, the check can't go here, as this method is expected to be infallible. It also can't go in the individual variants, as those don't have the visibility needed to do the check.

An alternative would be to hoist `word` to be settable on the enum itself, so you'd write `#[darling(word = Sit)]`; that has the advantage of being explicit that there can only be one variant which has this behavior, but it feels "wrong" somehow to me.

core/src/codegen/from_meta_impl.rs
```diff
@@ -90,16 +89,37 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                     }
                 };
 
+                let default_or_err = base.default.as_ref().map_or(
```
TedDriggs, line 92: Nit: I'd write this as a `map` call followed by an `unwrap_or_else` call, so that it's easier to see we transform `default_expr` if set, and then unwrap or create the error.

core/src/codegen/from_meta_impl.rs
```diff
@@ -90,16 +89,37 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
...
+                    .find_map(|variant| {
+                        if variant.word {
+                            let ty_ident = variant.ty_ident;
+                            let variant_ident = variant.variant_ident;
+                            Some(quote!(::darling::export::Ok(#ty_ident::#variant_ident)))
+                        } else {
+                            None
+                        }
+                    })
+                    .unwrap_or(quote!(::darling::export::Err(
```
TedDriggs, line 108: Nit: This should probably be `unwrap_or_else` to avoid an unnecessary invocation of `quote`.

core/src/codegen/from_meta_impl.rs
davidsemakula: I agree that hoisting to the enum feels "wrong" 🤔.

I think there should also be an error when a variant is annotated with both `skip` and `word`, does that make sense?

TedDriggs: I could actually see a use-case for using `skip` with `word` - it would let you make a variant that was only addressable in word form.

core/src/codegen/from_meta_impl.rs
```diff
@@ -90,16 +89,42 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
...
+                            ::darling::Error::unsupported_format("word")
+                        ))
+                    });
+
                 quote!(
                     fn from_list(__outer: &[::darling::export::NestedMeta]) -> ::darling::Result<Self> {
                         // An enum must have exactly one value inside the parentheses if it's not a unit
-                        // match arm
+                        // match arm.
+                        // NOTE: The empty list case uses the annotated default value for the enum if any.
```
TedDriggs, line 121: I'm not sure I agree with this behavior. Looking at `derive_buider` as an example, we see a difference in behavior between the "absent" case and the word-only case. If we get a field name with no contents, I'd expect that to error unless there was a `#[darling(word)]` variant present.

core/src/options/input_variant.rs
```diff
@@ -95,6 +99,22 @@ impl ParseAttribute for InputVariant {
...
 
             self.skip = FromMeta::from_meta(mi)?;
+        } else if path.is_ident("word") {
+            if self.word.is_some() {
+                return Err(Error::duplicate_field_path(path).with_span(mi));
+            }
+
+            if !self.data.is_unit() {
+                let note = "`#[darling(word)]` can only be applied to a unit variant";
+                #[cfg(feature = "diagnostics")]
```
TedDriggs, line 109: 👏 👏 👏  Props for taking this feature into consideration!

core/src/options/from_meta.rs
```diff
@@ -33,6 +35,26 @@ impl ParseData for FromMetaOptions {
...
+            Data::Enum(ref data) => {
+                // Adds errors for duplicate `#[darling(word)]` annotations across all variants.
+                let word_variants: Vec<_> = data
+                    .iter()
+                    .filter(|variant| variant.word.is_some())
+                    .collect();
+                if word_variants.len() > 1 {
+                    for variant in word_variants[1..].iter() {
+                        if let Some(word) = variant.word {
+                            errors.push(Error::duplicate_field("word").with_span(&word.span()));
```
TedDriggs, line 50: Rather than `duplicate_field` here, let's use a custom error:

```suggestion
                            errors.push(Error::custom("`word` can only be applied to one variant").with_span(&word.span()));
```

core/src/options/from_meta.rs
```diff
@@ -33,6 +35,26 @@ impl ParseData for FromMetaOptions {
...
+    fn validate_body(&self, errors: &mut Accumulator) {
+        match self.base.data {
+            Data::Enum(ref data) => {
+                // Adds errors for duplicate `#[darling(word)]` annotations across all variants.
+                let word_variants: Vec<_> = data
+                    .iter()
+                    .filter(|variant| variant.word.is_some())
+                    .collect();
+                if word_variants.len() > 1 {
+                    for variant in word_variants[1..].iter() {
```
TedDriggs, line 48: Let's apply the error to every usage, rather than just later ones.
```suggestion
                    for variant in &word_variants {
```

tests/compile-fail/duplicate_word_accross_variants.rs
```diff
@@ -0,0 +1,12 @@
+use darling::FromMeta;
```
TedDriggs, line 1: This file name has a typo in it: "accross" should be "across"

tests/compile-fail/duplicate_word_accross_variants.rs
```diff
@@ -0,0 +1,12 @@
...
+
+#[derive(Debug, FromMeta, PartialEq, Eq)]
```
TedDriggs, line 3: Nit: We only need the `FromMeta` derivation here, right?

TedDriggs (changes requested):
I am very excited about where this is heading; thank you for the latest updates. Feedback is now starting to get more specific, so I think we're closing in on readiness to merge.

davidsemakula:
LMK if I should squash the commits 🙂

core/src/options/input_variant.rs
```diff
@@ -95,6 +99,22 @@ impl ParseAttribute for InputVariant {
             }
 
             self.skip = FromMeta::from_meta(mi)?;
+        } else if path.is_ident("word") {
+            if self.word.is_some() {
+                return Err(Error::duplicate_field_path(path).with_span(mi));
+            }
+
+            if !self.data.is_unit() {
+                let note = "`#[darling(word)]` can only be applied to a unit variant";
```
TedDriggs, line 108: Nit: In the other error message, we didn't include the `#[darling(...)]` shell; should we also exclude it here? I forget what serde and the standard library do, so if you've got an example to help break the tie one way or the other that would be handy.

TedDriggs (commented):
The only remaining items are:

1. Deciding if this is the best name for this attribute. I thought it was originally (and I still don't think `default` is appropriate), but I am now worrying that `word` is too cryptic.
2. Updating the CHANGELOG
3. Updating the README

TedDriggs:
> LMK if I should squash the commits 🙂

I think we're very close to the squash point.

davidsemakula:
I'll work on the changelog and readme updates tomorrow.

> 1. Deciding if this is the best name for this attribute. I thought it was originally (and I still don't think `default` is appropriate), but I am now worrying that `word` is too cryptic.

Agreed, `default` would be confusing and `word` is a bit cryptic, but I don't have a better suggestion at this point. May be the more verbose `from_word`?

davidsemakula:
I updated the changelog. But I don't think I can update the readme to introduce and describe `#[darling(word)]` succinctly, can you take that part? I'm guessing it should be a bullet point under Features?

TedDriggs:
In the absence of a README entry, we could consider an example that includes several features of a heterogeneous enum. Maybe replicate one of the tests over to the examples and add some comments to show what's going on?

davidsemakula:
Sounds good, I'll work on an example 👍

davidsemakula:
> In the absence of a README entry, we could consider an example that includes several features of a heterogeneous enum. Maybe replicate one of the tests over to the examples and add some comments to show what's going on?

Added an example.

davidsemakula:
Pinging @TedDriggs

core/src/options/from_meta.rs
```diff
@@ -33,6 +35,31 @@ impl ParseData for FromMetaOptions {
     fn parse_field(&mut self, field: &syn::Field) -> Result<()> {
         self.base.parse_field(field)
     }
+
+    fn validate_body(&self, errors: &mut Accumulator) {
+        match self.base.data {
```
TedDriggs, line 40: Nit: I'd use `if let Data::Enum(ref data) = self.base.data {` here, rather than a `match` with an empty arm.

core/src/options/from_meta.rs
```diff
@@ -33,6 +35,31 @@ impl ParseData for FromMetaOptions {
...
+                        if let Some(word) = variant.word {
+                            errors.push(
+                                Error::custom(
+                                    "`#[darling(word)]` can only be applied to one variant",
+                                )
+                                .with_span(&word.span()),
+                            );
+                        }
+                    }
+                }
```
TedDriggs, line 58:
```suggestion
                let word_variants: Vec<_> = data
                    .iter()
                    .filter_map(|variant| variant.word.as_ref())
                    .collect();
                if word_variants.len() > 1 {
                    for word in word_variants {
                            errors.push(
                                Error::custom(
                                    "`#[darling(word)]` can only be applied to one variant",
                                )
                                .with_span(&word.span()),
                            );
                    }
                }
```

You can use `filter_map` here to avoid the later `if-let`

core/src/options/input_variant.rs
```diff
@@ -11,6 +12,7 @@ pub struct InputVariant {
     attr_name: Option<String>,
     data: Fields<InputField>,
     skip: Option<bool>,
+    pub word: Option<SpannedValue<bool>>,
```
TedDriggs, line 15:
```suggestion
    /// Whether or not the variant should be used to create an instance for
    /// `FromMeta::from_word`.
    pub word: Option<SpannedValue<bool>>,
```

Copying the doc comment over from the codegen field so people who are reading the `options` module to see what darling options are available can figure out why this exists.

core/src/options/mod.rs
```diff
@@ -146,4 +149,10 @@ pub trait ParseData: Sized {
     fn parse_field(&mut self, field: &syn::Field) -> Result<()> {
         Err(Error::unsupported_format("struct field").with_span(field))
     }
+
+    /// Perform validation checks that require data from more than one field or variant.
+    /// This default implementation is essentially a noop.
```
TedDriggs, line 154:
```suggestion
    /// The default implementation does no validations.
```

TedDriggs (approved):
Very minor changes left, then we're ready to merge!

Apologies for the delayed review, I was traveling last week and getting caught up at work this week.

TedDriggs (approved):

davidsemakula:
@TedDriggs can we get this published on crates.io when you have some time?
If I can do anything to help that process along, LMK 🙂

TedDriggs:
Will do on Monday; out of town this weekend
________________________________

davidsemakula:
Awesome! Thanks for taking the time to review this PR 🙂

PR #263, opened 2024-02-08
Fix some missing syn invisible group handling in FromMeta implementations

Ten0:
This was supposed to be fixed by c03616225175831db83a0b84b5866253f4d8ab43 but it looks like these implementations were forgotten.


TedDriggs (changes requested):
Can you please add some unit tests to `src/from_meta` that show inputs which rely on this change? That will make it easier to understand the change, and to prevent any future regressions.

Ten0:
As of syn 2.40, syn may generate this invisible token when the input to the darling proc macro (specifically, the attributes) are generated by a `macro_rules!` (e.g. propagating a `macro_rules!`'s `expr`), so this doesn't fit well within the existing test framework (as there would have to be tests for a crate that expands a such macro, that depends on the proc_macro user crate that depends on darling's macro crate).
However I did test that this resolves the parsing issue in such cases (We're actually relying on a fork for this ATM).
A similar breakage happened within Diesel: https://github.com/diesel-rs/diesel/pull/3873

TedDriggs:
Okay, can you add that as a comment in the source code with each of these changes so that it's clear later why this is going on?

TedDriggs (approved):

PR #265, opened 2024-02-08
Add FromMeta for Vec<LitStr> & co

Ten0:
This allows parsing meta attributes of the form `#[foo(bar = ["a", "b", "c"])]` to `Vec<syn::LitStr>` (and same for other `Lit`s).

### DIFF
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -1070,4 +1070,15 @@ mod tests {
             vec![0x50, 0xffffffff]
         );
     }
+
+    #[test]
+    fn test_lit_array() {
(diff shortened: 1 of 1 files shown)


TedDriggs (commented):
Why are these impls needed - wouldn't these all be valid `Expr`, or `ExprArray`? I'm not opposed to merging this, but I do recall there being some issue with `Vec` in the past, and I'm therefore wary of piecemeal adding `Vec` impls.

_Update:_ I think I remember the backstory now.

I avoided adding a `impl<T: FromMeta> FromMeta for Vec<T>` because different lists would have different manifestations in the AST. Adding these specific impls is useful because it lets someone take a list of homogenous literals, rather than having to accept an `Expr` and then check it using `and_then`

I am okay with merging this once the format errors are fixed.

PR #274, opened 2024-02-23
Support `with` on `attrs` magic field

### LINKED ISSUE #273: Support `with` on magic field `attrs`
In colin-kiegel/rust-derive-builder#310, @dtolnay pointed out some clunky code that handles “distribution and unnesting” of attributes. The `Options` struct is forced to have an `attrs` field that is only temporarily used.

This can be avoided by allowing the deriver of `FromDeriveInput` _et alia_ to provide a conversion function that takes attributes and returns a `darling::Result<?>`

This function must participate in the error accumulator behavior of the rest of the derived impl, so it must be invoked before the error check, and we’ll need a new local variable for the output.

`with` makes more sense than the postfix items, since this must allow failure and doesn’t itself take in a `Result`

TedDriggs:
Fixes #273


core/src/codegen/attr_extractor.rs
```diff
@@ -14,6 +15,9 @@ pub trait ExtractAttribute {
 
     fn forwarded_attrs(&self) -> Option<&ForwardAttrs>;
 
+    /// Gets the field that will receive forwarded attributes.
+    fn attrs_field(&self) -> Option<&AttrsField>;
```
TedDriggs, line 19: I think it might make sense to combine `ForwardAttrs` and `AttrsField` into one struct that represents both the forwarding policy and the receiving field.

TedDriggs: This has now been done.

PR #277, opened 2024-02-29
feat: add parse_attr_into

cathaysia:
add parse_attr_into macro, making the following code possible: 

```rust
#[derive(Debug, Default, FromMeta)]
#[darling(default)]
pub struct Attr {
    pub default: bool,
}

#[proc_macro_attribute]
pub fn api(attr: TokenStream, input: TokenStream) -> TokenStream {
    let _attr = parse_attr_into!(attr as Attr);
    input
}
```

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -40,3 +40,20 @@ pub use self::from_variant::FromVariant;
 pub use quote::ToTokens;
 #[doc(hidden)]
 pub use syn;
+
+#[macro_export]
+macro_rules! parse_attr_into {
+    ($attr:ident as $ty:ty) => {
+        match $crate::ast::NestedMeta::parse_meta_list($attr.into()) {
+            Ok(v) => match <$ty as $crate::FromMeta>::from_list(&v) {
+                Ok(v) => v,
+                Err(e) => {
+                    return TokenStream::from(e.write_errors());
+                }
+            },
+            Err(e) => {
+                return TokenStream::from($crate::Error::from(e).write_errors());
+            }
+        }
+    };
+}
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -41,19 +41,45 @@ pub use quote::ToTokens;
 #[doc(hidden)]
 pub use syn;
 
+/// Parse the attr TokenStream of a macro, triggering a compile error if the tokens fail to parse.
+/// # Example
+/// ```
+/// # extern crate proc_macro;
+/// # use proc_macro::TokenStream;
+/// # use darling::FromMeta;
+/// # use darling::parse_meta;
+///
+/// #[derive(Debug, Default, FromMeta)]
+/// #[darling(default)]
+/// pub struct Attr {
+///     pub default: bool,
+/// }
+///
+/// # const IGNORE: &str = stringify! {
+/// #[proc_macro_attribute]
+/// # };
+/// pub fn parse_attr(attr: TokenStream, input: TokenStream) -> TokenStream {
+///     let _attr = parse_meta!(attr as Attr);
+///     input
+/// }
+/// ```
+
 #[macro_export]
-macro_rules! parse_attr_into {
+macro_rules! parse_meta {
     ($attr:ident as $ty:ty) => {
         match $crate::ast::NestedMeta::parse_meta_list($attr.into()) {
             Ok(v) => match <$ty as $crate::FromMeta>::from_list(&v) {
                 Ok(v) => v,
                 Err(e) => {
-                    return TokenStream::from(e.write_errors());
+                    return proc_macro::TokenStream::from(e.write_errors());
                 }
             },
             Err(e) => {
-                return TokenStream::from($crate::Error::from(e).write_errors());
+                return proc_macro::TokenStream::from($crate::Error::from(e).write_errors());
             }
         }
+    };
+    ($attr:ident) => {
+        $crate::parse_meta!($attr as _)
     };
 }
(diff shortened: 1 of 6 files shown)


core/src/lib.rs
```diff
@@ -40,3 +40,20 @@ pub use self::from_variant::FromVariant;
 pub use quote::ToTokens;
 #[doc(hidden)]
 pub use syn;
+
+#[macro_export]
+macro_rules! parse_attr_into {
```
TedDriggs, line 45: Would `parse_meta` be a better name for this? While this macro would be helpful for attr macros, it expects a `TokenStream` containing a `NestedMeta` as input, rather than a full attribute.

core/src/lib.rs
```diff
@@ -40,3 +40,20 @@ pub use self::from_variant::FromVariant;
...
+    ($attr:ident as $ty:ty) => {
```
TedDriggs, line 46: I'd suggest making one branch with `as $ty:ty` and one without, as seen in [`syn::parse_macro_input`](https://docs.rs/syn/latest/syn/macro.parse_macro_input.html) - in the case where `as $ty` isn't provided, you can use `$crate::FromMeta::from_list` and expect Rust to figure out the type based on what's expected.

core/src/lib.rs
```diff
@@ -40,3 +40,20 @@ pub use self::from_variant::FromVariant;
...
+        match $crate::ast::NestedMeta::parse_meta_list($attr.into()) {
+            Ok(v) => match <$ty as $crate::FromMeta>::from_list(&v) {
+                Ok(v) => v,
+                Err(e) => {
+                    return TokenStream::from(e.write_errors());
```
TedDriggs, line 51: Should this use of `TokenStream` be fully-qualified like the other paths are?

core/src/lib.rs
```diff
@@ -40,3 +40,20 @@ pub use self::from_variant::FromVariant;
 pub use quote::ToTokens;
 #[doc(hidden)]
 pub use syn;
+
+#[macro_export]
```
TedDriggs, line 44: Please add a doc comment. Your example from the PR description was great, so I'd suggest including that in the doc comment after the explanation of the macro.

core/src/lib.rs
```diff
@@ -40,3 +40,20 @@ pub use self::from_variant::FromVariant;
...
+                Err(e) => {
+                    return TokenStream::from(e.write_errors());
+                }
+            },
+            Err(e) => {
+                return TokenStream::from($crate::Error::from(e).write_errors());
+            }
+        }
+    };
+}
```
TedDriggs, line 59: This commit should update [this example in README.md](https://github.com/TedDriggs/darling?tab=readme-ov-file#attribute-macros) to use the new macro.

(This has nothing to do with the current line, I just couldn't add the review comment to the README)

core/src/lib.rs
```diff
@@ -40,3 +40,20 @@ pub use self::from_variant::FromVariant;
...
+    ($attr:ident as $ty:ty) => {
+        match $crate::ast::NestedMeta::parse_meta_list($attr.into()) {
+            Ok(v) => match <$ty as $crate::FromMeta>::from_list(&v) {
+                Ok(v) => v,
+                Err(e) => {
+                    return TokenStream::from(e.write_errors());
+                }
+            },
+            Err(e) => {
+                return TokenStream::from($crate::Error::from(e).write_errors());
```
TedDriggs, line 55: Should this use of `TokenStream` be fully-qualified like the other paths are?

core/src/lib.rs
TedDriggs, line 59: Should there be some tests for this? The happy path can be normal tests, while invalid usages can go in the `compile-fail` folder so we're testing the error messages in the failure cases.

TedDriggs (changes requested):
Directionally this looks great - I've left some requests for documentation and testing, but the logic itself seems sound.

core/src/lib.rs
cathaysia: I add a doc test for this. But what's the `invalid usages`, can you give me some advice?

TedDriggs:
I made some changes [here](https://github.com/TedDriggs/darling/tree/feature/parse_attr) but for some reason those aren't showing up on the PR.

core/src/lib.rs
TedDriggs: See the branch I pushed (linked below); it includes those tests.

cathaysia:
I agree with your changes. Now that the changes have been written to the branch of this repository, this PR should be closed. If you think it's okay, please close this PR

cathaysia:
I see that [the branch](https://github.com/TedDriggs/darling/tree/feature/parse_attr) has not been merged yet. Do you have any questions?

TedDriggs:
No - I’m traveling for work right now, so haven’t had time to tidy this up.

TedDriggs:
I think it probably wants a `tt` or something else in this position; do you know what `syn`'s equivalent macro takes?

TedDriggs:
If we can revert and fix in rapid succession that'd be great, because I do like the concept here.

TedDriggs:
Yep, it's a `tt` ([source](https://docs.rs/syn/latest/syn/macro.parse_quote.html))

cathaysia:
may it should be `($id:ident $params:tt)`. Judging from the source code of quote, this piece needs to be parsed recursively.

TedDriggs:
Okay - I'm not sure how we reopen a PR; can you take a fork from my branch for this and submit a new PR with the updates off that?

cathaysia:
emm, ok.  I will do it tomorrow
________________________________

PR #279, opened 2024-03-12
Feature/parse attr

cathaysia:
continue of #277

NOTE: I add the permission, so you can push commits to cathaysia:feature/parse_attr, then it will autoupdate to this pr.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,10 +67,9 @@ This will produce a normal `darling::Result<T>` that can be used the same as a r
 ## Macro Code
 
 ```rust,ignore
-use darling::{Error, FromMeta, parse_meta};
-use darling::ast::NestedMeta;
-use syn::ItemFn;
+use darling::{parse_meta, FromMeta};
 use proc_macro::TokenStream;
+use syn::ItemFn;
 
 #[derive(Debug, FromMeta)]
 struct MacroArgs {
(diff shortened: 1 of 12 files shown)


cathaysia:
In the previous pr, you add a comment: https://github.com/TedDriggs/darling/pull/277#issuecomment-1988424260, I don't quite understand what you mean. What function do you want to achieve, or what is wrong with the current implementation?

TedDriggs:
I expect this to pass: `let x: NegatableIdentList = parse_meta!(inner(example::flag, my_feature));` - it didn't, which seems to be an issue with the macro construction where commas end the `tt`.

cathaysia:
Can you give a more detailed, working example? (Can be used to test parse_meta)

TedDriggs:
If I could give you a working example, then the PR wouldn't need amendment. Try using the macro with anything that contains a comma, and you get an error because the `tt` stops consuming at that point.

cathaysia:
I don't quite understand the purpose of this way of writing. What is inner(example::flag, my_feature)? Is it a function that returns a TokenStream?

TedDriggs:
Oh, wow, I'm sorry - this is entirely on me. I was mistakenly thinking of this as `parse_quote!` and was trying to write a fragment of a `Meta` that I expected to work here. Sorry about that. Let me have another look at this today, I think it's working properly and I'm the one who was off-base.

cathaysia:
This feature merging cycle far exceeded my expectations. On the one hand, considering that no obvious progress has been seen yet, on the other hand, this function has been completed. So I closed this pr. If you are still interested in this macro. You can implement it on your own branch.

PR #281, opened 2024-04-01
Bump to edition 2021

GnomedDev:
This crate is still on edition 2018 but it's MSRV is exactly the MSRV needed for 2021... so let's get this over with. It turns out that there were no changes required, but I fixed all the things missed for the *2018* migration by adding `#![warn(rust_2018_idioms)]`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ implementing custom derives.
 """
 license = "MIT"
 readme = "README.md"
-edition = "2018"
+edition = "2021"
 exclude = ["/.travis.yml", "/publish.sh", "/.github/**"]
 
 [badges]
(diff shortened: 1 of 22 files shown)


TedDriggs:
It looks like most of the code change here is due to the [`elided_lifetimes_in_paths`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#elided-lifetimes-in-paths) lint, which says:

> This lint is "allow" by default because it has some known issues, and may require a significant transition for old code.

Looking around further, it seems that use of that specific lint is not universally popular ([source](https://github.com/dtolnay/cxx/issues/907#issuecomment-991313362)).

I'm otherwise supportive of this PR.

What would your thoughts be on allowing that one lint, and otherwise warning on 2018 idioms?

GnomedDev:
I personally would take dtolnay's opinions with a grain of salt, as he opposes a lot of common guidelines. I have never seen a codebase which does not follow this lint and is using rust_2018_idioms, so would be quite opposed to not following common guideline.

TedDriggs:
You’re right… it just looks uglier ☹️

TedDriggs (approved):

PR #282, opened 2024-04-01
Fix CI by bumping compiletest rust version

GnomedDev:


TedDriggs:
The fixed version of rustc is necessary to avoid CI failures due to compiler error messages changing across versions. What’s the goal of removing it here?

GnomedDev:
I figured that out before you commented that, and have already changed the PR to simple bump the version up to latest.

TedDriggs (approved):

PR #284, opened 2024-04-09
Fix `cargo test --no-default-features`

mgeisler:
The suggestion is not given when the `suggestions` Cargo feature is disabled.

### DIFF
--- a/tests/flatten.rs
+++ b/tests/flatten.rs
@@ -165,7 +165,7 @@ fn do_not_suggest_invalid_alts() {
 }
 
 #[test]
-#[cfg(features = "suggestions")]
+#[cfg(feature = "suggestions")]
 fn suggest_valid_parent_alts() {
     let errors = Outer::from_derive_input(&parse_quote! {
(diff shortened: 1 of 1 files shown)


tests/flatten.rs
```diff
@@ -165,6 +165,7 @@ fn do_not_suggest_invalid_alts() {
 }
 
 #[test]
+#[cfg(features = "suggestions")]
```
TedDriggs, line 168:
```suggestion
#[cfg(feature = "suggestions")]
```

Per clippy, this is invalid as-written.

TedDriggs (changes requested):

TedDriggs (approved):

TedDriggs:
Now the one open question in my mind - why didn't CI catch this before? Don't need to block this PR for it, but do need to sort that out.

mgeisler:
> Now the one open question in my mind - why didn't CI catch this before? Don't need to block this PR for it, but do need to sort that out.

I think it's because you're not running all tests with `--no-default-features`, only the ones in `core/` if I read [`ci.yml`](https://github.com/TedDriggs/darling/blob/master/.github/workflows/ci.yml#L36) correctly

PR #287, opened 2024-04-18
feat: Add `FromMeta` support for `BTreeMap`

Techassi:
This PR adds an `FromMeta` implementation for `BTreeMap<String, V>` and `BTreeMap<syn::Ident, V>` by adjusting the existing macro. The macro can now handle both `HashMap` and `BTreeMap`, because much of the underlying code is the same.

I also added new unit tests to test the `BTreeMap` implementation.

### DIFF
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::cell::RefCell;
+use std::collections::btree_map::BTreeMap;
 use std::collections::hash_map::HashMap;
 use std::collections::HashSet;
 use std::hash::BuildHasher;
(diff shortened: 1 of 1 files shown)


TedDriggs:
The code for this looks fine to me, but I am curious what the use-case is that needs this. Are you doing something that needs the keys to always be sorted?

Techassi:
Yeah I had a very rare use-case where this behaviour was required. But I can see that this feature seems to be needed very rarely.

TedDriggs:
@Techassi is this something you'd still use if it was merged? If so, can you elaborate on the use-case?

Techassi:
One specific use-case was the definition of versions (like SemVer or Kubernetes API versions), which would enforce proper ordering via (Partial)Ord.

But I can see that this feature would be used very rarely, so feel free to close this PR if it introduces too much complexity.

Veetaha:
I think this feature would be rather useful. BTreeMap is underrated. The main thing is that sorting guarantee gives *stability*. I personally prefer `BTree` data structures due to that. It helps guarding against flaky bugs. One case from production was taking the first element from a map and deciding which DB index to use based on that element. By using a HashMap that decisions was always doing a random choice of an element and thus introduced a flaky bug that reproduced randomly. I believe same concerns apply anywhere, where the iteration order of the map may influence the behavior of the program, including a proc macro.

So having a `Btree` variant of data structure support is a win, and would be cool to get this merged.

TedDriggs:
I will review and merge next week.

core/src/from_meta.rs
```diff
@@ -693,89 +694,106 @@ impl KeyFromPath for syn::Ident {
...
+                HashMap::with_capacity_and_hasher($nested.len(), Default::default()),
+                $key,
+                $nested
+            );
+        }
+    };
 
-                        let already_seen = seen_keys.contains(&key);
+    (btree_map, $key:ty, $nested:ident) => {
+        impl<V: FromMeta + Ord> FromMeta for BTreeMap<$key, V> {
```
TedDriggs, line 709: Why does `V` have the `Ord` constraint here?

TedDriggs: I tested this, and found that the `Ord` constraint was unnecessary and created problems. I've removed it, and added a test for a non-`Ord` value.

PR #288, opened 2024-05-10
Add support for non-zero integers.

mpapierski:
This PR adds support for `NonZero` types from the standard library. I needed a way to perform some basic validation of a user input and ensure that the passed value is not a zero.

### DIFF
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -3,6 +3,10 @@ use std::cell::RefCell;
 use std::collections::hash_map::HashMap;
 use std::collections::HashSet;
 use std::hash::BuildHasher;
+use std::num::{
+    NonZeroI128, NonZeroI16, NonZeroI32, NonZeroI64, NonZeroI8, NonZeroIsize, NonZeroU128,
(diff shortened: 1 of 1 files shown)


core/src/from_meta.rs
```diff
@@ -795,8 +813,12 @@ mod tests {
 
     #[track_caller]
     fn fm<T: FromMeta>(tokens: TokenStream) -> T {
+        try_fm(tokens).expect("Tests should pass valid input")
+    }
+
+    #[track_caller]
+    fn try_fm<T: FromMeta>(tokens: TokenStream) -> Result<T> {
```
TedDriggs, line 820: Rather than this, could you use `#[should_panic]` on the failing test?

TedDriggs (changes requested):

core/src/from_meta.rs
mpapierski: Sure thing, I just pushed this change in 47f9a4c230001f42178d9c6d89e620268abe90ed

TedDriggs (approved):

PR #292, opened 2024-05-30
Emit unwrap_or_default when appropriate for default fields

novafacing:
Clippy has recently started emitting error messages like:

```
warning: if let can be simplified with `.unwrap_or_default()`
  --> simics-macro/src/init/mod.rs:16:15
   |
16 |     #[darling(default)]
   |               ^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_unwrap_or_default
help: replace it with
   |
16 ~     #[darling(default)]
17 +     /// The name of the module, used in the `MOD:NAME` expression
18 ~     pub name.unwrap_or_default())]
```

This PR just uses `unwrap_or_default` as Clippy wants and the error message goes away.

### DIFF
--- a/core/src/codegen/field.rs
+++ b/core/src/codegen/field.rs
@@ -224,15 +224,11 @@ impl<'a> ToTokens for Initializer<'a> {
                 quote!(#ident: #ident)
             }
         } else if let Some(ref expr) = field.default_expression {
-            if matches!(expr, DefaultExpression::Trait { .. }) {
-                quote_spanned!(expr.span()=> #ident: #ident.1.unwrap_or_default())
+            quote_spanned!(expr.span()=> #ident: if let Some(__val) = #ident.1 {
+                __val
             } else {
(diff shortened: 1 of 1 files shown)


TedDriggs:
@novafacing it appears that this change has broken one of the compile tests. I'm not thrilled with the change in the compiler diagnostic - maybe adding an `#[allow(clippy::...)]` is the better move?

novafacing:
That works for me, I don't really like this diagnostic either. I updated the PR to just add that `allow`, if you're cool with that.

martin-g:
Related to https://github.com/rust-lang/rust-clippy/issues/12643

TedDriggs:
This is being superseded by #296, which puts the `allow` on the generated trait impl rather than trying to hunt down each place where it appears.

I don't have the time right now, but it would be nice if there was a clippy issue to disable this check in blocks that have `#[automatically_derived]` on them; this seems like a stylistic recommendation that doesn't make sense to emit on code that may be generic.

PR #310, opened 2024-10-02
Accept a closure for `with` in lieu of a path for fields

### LINKED ISSUE #309: Accept an arbitrary expression that returns a callable object in `#[darling(with)]`
It's not a big issue, but just an inconvenience. Here is an example of this problem that [I have](https://github.com/elastio/bon/blob/eb390012783b7fcf6f68223a30965b81b2b50f5a/bon-macros/src/builder/builder_gen/builder_params.rs#L9-L33).

I need to configure parsing with some parameterization necessary for validation and error messages. Currently to do that you always need to define a separate function item that accepts a `&syn::Meta` and returns a `Result`. It would be better if I could just write an expression that returns a closure to invoke in the `#[darling(with)]` directly like this:

```rust
use darling::FromMeta;

#[derive(FromMeta)]
struct Example {
    #[darling(with = parse_item_params(... config here ...))]
    finish_fn: ItemParams,
   
    // alternatively just pass a closure 
    #[darling(with = |meta| parse_item_params(meta, ... config here ...))]
    builder_type: ItemParams,
}

fn parse_item_params(config: ItemParamsParsing) -> impl FnOnce(&syn::Meta) -> Result<ItemParams> { /**/ }
fn parse_item_params_alt(meta: &syn::Meta, config: ItemParamsParsing) -> Result<ItemParams> { /**/ }
```

The main goal is to provide additional context for the parser function passed to `with`.

TedDriggs:
Fixes #309

This has some issues, so it's not yet ready to merge:

1. It doesn't work for all the places where `with` is used.
2. I'm not thrilled with the introduction of new locals to support this, and am wondering if there's a way to write it that better encapsulates the desired behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 -  Support `#[darling(with = ...)]` on the `data` field when deriving `FromDeriveInput`. This allows the use of simpler receiver types, such as a `Vec` of enum variants.
 -  Bump version of `proc-macro2` to 1.0.86.
+-  Accept closures for `#[darling(with = ...)]` on fields in `FromDeriveInput`, `FromMeta`, `FromField`, etc. [#309](https://github.com/TedDriggs/darling/issues/309)
 
 ## v0.20.10 (July 9, 2024)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ syn = "2.0.15"
 
 [target.'cfg(compiletests)'.dev-dependencies]
 rustversion = "1.0.9"
-trybuild = "1.0.99"
+trybuild = "1.0.89"
 
 [features]
 default = ["suggestions"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ syn = "2.0.15"
 
 [target.'cfg(compiletests)'.dev-dependencies]
 rustversion = "1.0.9"
-trybuild = "1.0.38"
+trybuild = "1.0.99"
 
 [features]
 default = ["suggestions"]
--- a/compiletests.sh
+++ b/compiletests.sh
@@ -1,1 +1,1 @@
-RUSTFLAGS="--cfg=compiletests" cargo +1.65.0 test --test compiletests
+RUSTFLAGS="--cfg=compiletests" cargo +1.77.0 test --test compiletests
--- a/core/src/codegen/field.rs
+++ b/core/src/codegen/field.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 
 use proc_macro2::TokenStream;
 use quote::{quote, quote_spanned, ToTokens, TokenStreamExt};
-use syn::{spanned::Spanned, Ident, Path, Type};
+use syn::{spanned::Spanned, Ident, Type};
 
 use crate::codegen::{DefaultExpression, PostfixTransform};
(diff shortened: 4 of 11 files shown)


Veetaha:
If you'd like to avoid using a local variable there are two ways to do that:

The naive way us to just cast the expression with `as`: `#closure as fn(...)`. Although the `as` operator is a bit more powerful for this task, because it's meant for potential type coersions, while in this case we just need a type hint.

Another way to do that is with `::core::convert::identity::<fn(...) -> _>(#closure)` which provides a hint for the expression

TedDriggs:
Using `identity` seems to work well. It also solves one of the main barriers to allowing arbitrary expressions in this position, though I'm still hesitant to allow arbitrary expressions until I can test the isolation and the quality of the error messages when the user does something wrong.

TedDriggs:
@Veetaha I've updated this using `core::convert::identity`, added a compile-fail test to show what happens if you try to access locals, and renamed `darling_core::codegen::Field::with_path` to be the more descriptive `with_callable`.

Veetaha (approved):
LGTM, except docs should need a small update

TedDriggs:
@Veetaha ugh, one more question for you: Any ideas on how to fix the output of the compiletest so it will pass both locally and on CI? It seems the issue is the path on the last line of the error, potentially due to the use of `$RUST` vs a specified path?

Veetaha:
I haven't used `compiletest`. I assume it's an older equivalent of [`trybuild`](https://docs.rs/trybuild/latest/trybuild/), which I've been using for my tests. Maybe there it will work with `trybuild`?

The version of trybuild, that supports 1.56 MSRV is `1.0.89` (can be found in the [table here](https://lib.rs/crates/trybuild/versions)).

TedDriggs:
I'm using `trybuild` - the `compiletests` is just the test directory name, since I didn't want to confuse it by recycling the name.

These tests are run on 1.77.0; I'm wondering if the issue is the lack of an environment variable for the Rust directory in the local `compiletests.sh` file, since it looks like the CI version of the test is using `$RUST` where the local version used that absolute path.

TedDriggs:
Okay, it looks like `trybuild@1.0.89` is the sweet spot where builds on 1.56.0 still work, though why CI is ever trying to build with `trybuild` on 1.56.0 is unclear to me, as that dependency is only present for `cfg(compiletests)` and that's only used on 1.77.0.

Veetaha:
>  why CI is ever trying to build with trybuild on 1.56.0 is unclear to me, as that dependency is only present for cfg(compiletests) and that's only used on 1.77.0

Do you mean this error?:

> failed to select a version for the requirement

If yes, then that's expected because this happens not at the building stage, but when updating the crates.io index. `cargo` needs to make sure the dependency exists (even if its under a cfg), and create a single `Cargo.lock` file which lists the pinned dependency versions assuming all dependencies are used

TedDriggs:
Yes, that's the error I meant. Thanks for the explanation. For now, it seems that leaving the dependency at 1.0.89 is sufficient. If it ever breaks, that'll probably be a sign it's time to bump `darling`'s MSRV for its next release.

I've updated examples, tests, and CHANGELOG for this feature; I didn't see any places in the actual docs where a change was needed. I do want to look at expanding this to cover the other `with` scenarios, though I don't think that needs to be part of this PR.

PR #313, opened 2024-11-16
Add `from_word` impl for structs with default

AdinAck:
Before, the `FromMeta` derive macro did not implement `from_word` for structs which are annotated with `#[darling(default)]`, even though it could.

So now expressions like:

`#[attr(arg)]`

`#[attr(arg(arg1, arg2))]`

are possible.

The former assumes the default value, and the latter allows for specification of fields.

### DIFF
--- a/core/src/codegen/from_meta_impl.rs
+++ b/core/src/codegen/from_meta_impl.rs
@@ -54,6 +54,14 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                 let default = base.fallback_decl();
                 let post_transform = base.post_transform_call();
 
+                let word_impl = base.default.as_ref().and_then(|_| {
+                    Some(quote! {
+                        fn from_word() -> ::darling::Result<Self> {
+                            Ok(::darling::export::Default::default())
+                        }
+                    })
+                });
+
                 quote!(
                     fn from_list(__items: &[::darling::export::NestedMeta]) -> ::darling::Result<Self> {
 
@@ -73,6 +81,8 @@ impl<'a> ToTokens for FromMetaImpl<'a> {
                             #(#inits),*
                         }) #post_transform
                     }
+
+                    #word_impl
                 )
             }
             Data::Enum(ref variants) => {
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -426,6 +426,7 @@ macro_rules! from_syn_expr_type {
 
 from_syn_expr_type!(syn::ExprArray, Array);
 from_syn_expr_type!(syn::ExprPath, Path);
+from_syn_expr_type!(syn::ExprRange, Range);
 
 /// Adapter from `syn::parse::Parse` to `FromMeta` for items that cannot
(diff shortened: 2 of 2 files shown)


TedDriggs:
I don't _think_ I like this, though I could be convinced otherwise.

I've seen a number of macros that have three behaviors:

* Property is absent (e.g. the whole setting is disabled)
* Property is present as a word (the setting is enabled, with default values)
* Property is present as a key-value pair or as a meta list (the setting is enabled and customized)

This proposal collapses the first two states, and I don't see a benefit to doing so. If the word and no-word states are equivalent, then removing the word is a valid solution to the error, resulting in cleaner source code.

TedDriggs:
I can see value in providing a way to specify a custom implementation of `from_word` at the struct level; that would allow someone to have it point to the default, but would also accommodate any of the word-or-list use-cases without needing to use `with` separately.

If we went that route, I'd expect that `#[darling(from_word = ...)]` is valid at the struct level and takes a path (there's a separate effort which would allow a closure in lieu of a path) to a function whose signature which is `fn() -> darling::Result<Self>`.

There'd be a small amount of extra typing for people who wanted to forward `from_word` to `Default::default`, but it's not much.

If you're interested in pursuing this, file an issue and we can work through some of the details (e.g. whether there are cases where using `from_word` is disallowed, whether we should also include an option for overriding other `from_*` methods, etc.)

AdinAck:
It's certainly possible my use case is... weird and this isn't the best way to approach this. I don't understand enough about the inner workings of darling to pursue what you proposed, nor the time, but I appreciate the effort.

Thanks!

TedDriggs:
Can you share a bit more about your use-case? Why would you have a struct that impls `Default` and want to let someone get the default value by including the word?

AdinAck:
> Can you share a bit more about your use-case?

Part of my project involves annotating registers in microcontrollers. There are many properties, read/write access being one of them.

A simple case is a register has read and write access:

```rust
#[register(offset = 0x1c, schema = whatever, read, write)]
mod some_register { ... }
```

A less simple case is when the read or write access have some additional properties/constraints (register is only read under certain conditions):

```rust
#[register(offset = 0x1c, schema = whatever, read(entitlements = [...], effects = [...]), write)]
mod some_register { ... }
```

This felt the most ergonomic to me, to just use the word when there are no special properties, and to refine the annotation with optional properties when need be as shown above.

TedDriggs:
Can you take a look at #323 and see if that would meet your needs?

AdinAck:
Ah you're saying: `#[darling(from_word = Default::default())]`?

TedDriggs:
Close! It'd need to be `#[darling(from_word = || Ok(Default::default()))]`, since it expects a `Result` from you. Happy to discuss the ergonomics of that on the open PR :)

AdinAck:
Ah, gotcha, no it makes sense. That's not too bad, it solves my problem and doesn't introduce the problems your were worried about. Sounds good to me!

PR #314, opened 2024-11-28
Make <LitStr as FromMeta>::from_string no longer return an error

Adhalianna:
The `FromMeta` implementation for `syn::LitStr` is now more complete fixing difficult to debug macro
panics that would occur when the type was used to build the macro.

Without this commit I would get the following panics:
```
error: proc-macro derive panicked
 --> models/src/lib.rs:1:44
  |
1 | #[derive(serde::Deserialize, Debug, Clone, jayapi::AsResource)]
  |                                            ^^^^^^^^^^^^^^^^^^
  |
  = help: message: called `Result::unwrap()` on an `Err` value: Error { kind: UnexpectedType("string"), locations: [], span: None }
```
for my derive macro built using darling. One of the structs nested on the one that had derived `darling::FromDeriveInput` used `syn::LitStr` type as its field. With this commit which actually adds only `from_string`  function, lines 579-581, (the rest is copy-paste from the `from_meta_lit!` macro) the panic is gone. 

I am not familiar with `syn` so I guess there might be a smarter way of creating LitStr from a &str but this works for me perfectly fine.

### DIFF
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -562,12 +562,49 @@ macro_rules! from_meta_lit {
 
 from_meta_lit!(syn::LitInt, Lit::Int);
 from_meta_lit!(syn::LitFloat, Lit::Float);
-from_meta_lit!(syn::LitStr, Lit::Str);
 from_meta_lit!(syn::LitByte, Lit::Byte);
 from_meta_lit!(syn::LitByteStr, Lit::ByteStr);
 from_meta_lit!(syn::LitChar, Lit::Char);
 from_meta_lit!(syn::LitBool, Lit::Bool);
 from_meta_lit!(proc_macro2::Literal, Lit::Verbatim);
+
+impl FromMeta for syn::LitStr {
+    fn from_value(value: &syn::Lit) -> Result<Self> {
+        if let syn::Lit::Str(ref value) = *value {
+            Ok(value.clone())
+        } else {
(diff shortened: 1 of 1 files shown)


TedDriggs:
How/why are you getting into the situation that this is being called with a `&str` while the receiver field wants a `LitStr`?

I'm torn on this, because the point of using a `LitStr` receiver field rather than a `String` would be to keep the span information, so we're adding a new impl in order to allow someone to wrap a `String` without that span information.

TedDriggs:
Closing for repo hygiene due to lack of response, but willing to continue this discussion.

TedDriggs:
> I also do not understand how just choosing syn::LitStr gives one access to span information.

Because `syn::LitStr` internally has a field for the `span` in the input token stream where that string literal was seen, and `String` doesn't.

> Ok, changing all those syn::LitStr into String turned out to be actually very easy and made the macros work but I am still confused why it is so wrong to use syn::LitStr. I do not remember entirely how I arrived at syn::LitStr back then but seeing how syn::Path is fine really makes me want to have syn::LitStr working too.

The piece I'm confused about here is why it wasn't working when you used `LitStr` - it's _supposed_ to work the way you described.

The only thing I can think of is that your `TryFrom` implementation is directly calling `from_string` rather than calling `from_meta`. That would cause the `Err` value you see in your unwrap. 

Can you look to see if there are **_any_** references to `FromMeta::from_string` in your code and share those?

PR #318, opened 2025-01-17
Error if `attrs` field is declared and `forward_attrs` is missing.

### LINKED ISSUE #317: Emit useful error when attrs field is used and forward_attrs is missing
I ran across an error ("Errors were already checked") using FromDeriveInput that seems a little difficult to understand and doesn't seem to have a good explanation anywhere. I narrowed it down to when args is included in the struct. Full repro:

```rust
use darling::FromDeriveInput;
use proc_macro::TokenStream;
use syn::{parse_macro_input, Attribute, Ident};

#[proc_macro_derive(Hello)]
pub fn derive_hello(input: TokenStream) -> TokenStream {
    let input = parse_macro_input!(input as syn::DeriveInput);
    let _args = HelloArgs::from_derive_input(&input);
    TokenStream::new()
}

#[derive(FromDeriveInput)]
struct HelloArgs {
    ident: Ident,
    attrs: Vec<Attribute>, // This is the problem
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_derive_hello() {
        let input = r#"
            #[derive(Hello)]
            struct World;
        "#;
        let input = syn::parse_str(input).unwrap();
        let args = HelloArgs::from_derive_input(&input).unwrap();
        assert_eq!(args.ident, "World");
    }
}
```

```
---- tests::test_derive_hello stdout ----
thread 'tests::test_derive_hello' panicked at src/lib.rs:12:10:
Errors were already checked
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: core::option::Option<T>::expect
             at /Users/joshka/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:933:21
   4: <darling_error::HelloArgs as darling_core::from_derive_input::FromDeriveInput>::from_derive_input
             at ./src/lib.rs:12:10
   5: darling_error::tests::test_derive_hello
             at ./src/lib.rs:30:20
   6: darling_error::tests::test_derive_hello::{{closure}}
             at ./src/lib.rs:24:27
   7: core::ops::function::FnOnce::call_once
             at /Users/joshka/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
   8: core::ops::function::FnOnce::call_once
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/ops/function.rs:250:5
```

I'm not sure if this is a bug in darling, I'm assuming that I'm probably using this wrong, but the error message doesn't make it easy to understand what do next. It's also problematic as it doesn't really point back to the `attrs` field as being the problem. Consider adding extra info about what this means and what is missing from the struct / config to make it correct, or documenting this in the docs (the only hit for this error message is the line of code where it's omitted).

TedDriggs:
Previously, this would result in a cryptic error due to an expect(...) failure if neither `attributes` nor `forward_attrs` were declared.

This change is technically breaking, as it was previously possible to declare this field without using `forward_attrs`; if `attributes` was used the `attrs` field would be initialized to an empty array. That behavior was more likely to mask issues than be useful, but it's possible someone declared the field in order to manually populate it later.

Fixes #317

### DIFF
--- a/core/src/options/outer_from.rs
+++ b/core/src/options/outer_from.rs
@@ -1,12 +1,13 @@
 use syn::spanned::Spanned;
 use syn::{Field, Ident, Meta};
 
+use crate::ast::Data;
 use crate::codegen::ForwardAttrs;
 use crate::options::{
     Core, DefaultExpression, ForwardAttrsFilter, ForwardedField, ParseAttribute, ParseData,
 };
 use crate::util::PathList;
-use crate::{FromField, FromMeta, Result};
+use crate::{Error, FromField, FromMeta, Result};
 
 /// Reusable base for `FromDeriveInput`, `FromVariant`, `FromField`, and other top-level
 /// `From*` traits.
@@ -91,5 +92,20 @@ impl ParseData for OuterFrom {
 
     fn validate_body(&self, errors: &mut crate::error::Accumulator) {
         self.container.validate_body(errors);
+        if let Some(attrs) = &self.attrs {
+            if self.forward_attrs.is_none() {
+                let container_name = match &self.container.data {
+                    Data::Enum(_) => "enum",
+                    Data::Struct(_) => "struct",
+                };
+                errors.push(
+                    Error::custom(format!(
+                        "field will not be populated because `forward_attrs` is not set on the {}",
+                        container_name
+                    ))
+                    .with_span(&attrs.ident),
+                );
+            }
+        }
     }
 }
--- a/tests/compile-fail/attrs_without_forward_attrs.rs
+++ b/tests/compile-fail/attrs_without_forward_attrs.rs
@@ -1,0 +1,10 @@
+use darling::FromDeriveInput;
+use syn::{Attribute, Ident};
+
+#[derive(FromDeriveInput)]
+struct HelloArgs {
+    ident: Ident,
+    attrs: Vec<Attribute>,
+}
+
+fn main() {}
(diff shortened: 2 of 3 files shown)


TedDriggs:
@joshka would this error have helped you diagnose your problem in #317 faster?

core/src/options/outer_from.rs
```diff
@@ -91,5 +92,17 @@ impl ParseData for OuterFrom {
 
     fn validate_body(&self, errors: &mut crate::error::Accumulator) {
         self.container.validate_body(errors);
+        if let Some(attrs) = &self.attrs {
+            if self.forward_attrs.is_none() {
+                let container_name = match &self.container.data {
+                    Data::Enum(_) => "enum",
+                    Data::Struct(_) => "struct",
+                };
```
TedDriggs, line 100: I'm surprised this hasn't been used before; might be worth adding a function to `Data` for it?

joshka: Perhaps check out [strum](https://crates.io/crates/strum) for a quick derive macro approach to giving enums display names

joshka (approved):
LGTM this would have helped point in the right direction.

core/src/options/outer_from.rs
TedDriggs: This isn't actually getting the name, it's just the word "enum" or "struct"

PR #323, opened 2025-01-22
Allow custom from_word and from_none behaviors in derived FromMeta impls

### LINKED ISSUE #320: Accept from_* methods as #[darling(...)] fields
It's currently impossible to have a struct with a derived `FromMeta` impl that has custom behavior in the `from_word` or `from_none` cases. The only workaround is to use `with` on every field of that type, which is tiresome.

Based on the discussion in #313, this proposal would allow the deriving struct to specify a path (or closure) that replaces the body of the corresponding method.

This will require some validation to ensure these options aren't conflicting with generated impls; for example, `from_word` cannot be used with enums where `word` is in use.

Additionally, this will require new documentation and examples.

TedDriggs:
Fixes #320

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,8 @@ Darling's features are built to work well for real-world projects.
 -   **Span access**: Use `darling::util::SpannedValue` in a struct to get access to that meta item's source code span. This can be used to emit warnings that point at a specific field from your proc macro. In addition, you can use `darling::Error::write_errors` to automatically get precise error location details in most cases.
 -   **"Did you mean" suggestions**: Compile errors from derived darling trait impls include suggestions for misspelled fields.
 -   **Struct flattening**: Use `#[darling(flatten)]` to remove one level of structure when presenting your meta item to users. Fields that are not known to the parent struct will be forwarded to the `flatten` field.
+-   **Custom shorthand**: Use `#[darling(from_word = ...)]` on a struct or enum to override how a simple word is interpreted. By default, it is an error for your macro's user to fail to specify the fields of your struct, but with this you can choose to instead produce a set of default values. This takes either a path or a closure whose signature matches `FromMeta::from_word`.
+-   **Custom handling for missing fields**: When a field is not present and `#[darling(default)]` is not used, derived impls will call `FromMeta::from_none` on that field's type to try and get the fallback value for the field. Usually, there is not a fallback value, so a missing field error is generated. `Option<T: FromMeta>` uses this to make options optional without requiring `#[darling(default)]` declarations, and structs and enums can use this themselves with `#[darling(from_none = ...)]`. This takes either a path or a closure whose signature matches `FromMeta::from_none`.
 
 ## Shape Validation
 
--- a/core/src/codegen/from_meta_impl.rs
+++ b/core/src/codegen/from_meta_impl.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use proc_macro2::TokenStream;
 use quote::{quote, quote_spanned, ToTokens};
 use syn::spanned::Spanned;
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -123,6 +123,8 @@ Darling's features are built to work well for real-world projects.
 -   **Span access**: Use `darling::util::SpannedValue` in a struct to get access to that meta item's source code span. This can be used to emit warnings that point at a specific field from your proc macro. In addition, you can use `darling::Error::write_errors` to automatically get precise error location details in most cases.
 -   **"Did you mean" suggestions**: Compile errors from derived darling trait impls include suggestions for misspelled fields.
 -   **Struct flattening**: Use `#[darling(flatten)]` to remove one level of structure when presenting your meta item to users. Fields that are not known to the parent struct will be forwarded to the `flatten` field.
+-   **Custom shorthand**: Use `#[darling(from_word = ...)]` on a struct or enum to override how a simple word is interpreted. By default, it is an error for your macro's user to fail to specify the fields of your struct, but with this you can choose to instead produce a set of default values. This takes either a path or a closure whose signature matches `FromMeta::from_word`.
```
TedDriggs, line 126: I considered not requiring the outer `Result`, as that would greatly increase the likelihood of path values "just working", and given that there are no inputs to this function it seems unlikely that someone would want to return an `Err`.

Unfortunately, I feel like this sort of common-path optimization has burned `darling` in the past, either in the form of increased complexity or in making certain concepts impossible to express; maybe the `from_word` function is calling a method on a type parameter and that can sometimes fail.

Given the signatures for these two methods, I expect the closure form will be more common here to handle the `Ok` and `Some` wrapping.

core/src/codegen/from_meta_impl.rs
```diff
@@ -89,22 +113,22 @@ impl ToTokens for FromMetaImpl<'_> {
                     }
                 };
 
-                let word_or_err = variants
+                let from_word_method = variants
                     .iter()
                     .find_map(|variant| {
```
TedDriggs, line 118: Not necessary right now, but might be worth exploring generating the `from_word` method in `FromMetaOptions` to simplify this code generation.

core/src/options/input_field.rs
```diff
@@ -149,7 +149,7 @@ impl ParseAttribute for InputField {
                 return Err(Error::duplicate_field_path(path).with_span(mi));
             }
 
-            self.with = Some(With::from_meta(mi)?);
+            self.with = Some(Callable::from_meta(mi)?);
```
TedDriggs, line 152: Nit: Other attributes are using `FromMeta::from_meta` instead of naming the target type. This should follow that pattern.

tests/from_meta.rs
```diff
@@ -107,3 +107,150 @@ mod enum_impl {
...
     }
 }
+
+mod from_none_struct_closure {
+    use darling::FromMeta;
+    use syn::parse_quote;
+
+    #[derive(Debug, FromMeta)]
+    struct Outer {
+        // Do NOT add `darling(default)` here.
```
TedDriggs, line 117: Explain why not.

tests/from_meta.rs
```diff
@@ -107,3 +107,150 @@ mod enum_impl {
...
+        assert_eq!(error.len(), 1);
+    }
+
+    #[test]
+    fn list_sets_field() {
+        let thing = Outer::from_list(&[parse_quote!(speech(max_volume = 5))]).unwrap();
+        assert_eq!(thing.speech.max_volume, 5);
+    }
+}
+
```
TedDriggs, line 146: Add `from_none_struct_path` for completeness.

core/src/codegen/from_meta_impl.rs
TedDriggs: I went ahead and did this; it does make the `codegen` piece simpler to read, which I like.

PR #341, opened 2025-05-15
feat: Add optional serde support for IdentString

Techassi:
This PR adds a new optional feature called `serde` which enables Deserialization and Serialization support for `IdentString`. Serde support can be extended in the future and is always opt-in.


Techassi:
It seems that `dep:` is only available since Rust 1.60. If I remove that, all CI checks break. Currently only the tests on Rust 1.56 break. Can you provide a hint on how to fix this, because it seems to work in general (e.g. the `suggestions` feature also uses an optional dependency).

core/src/serde.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+
+impl serde::Serialize for IdentString {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(self.as_str())
+    }
```
martin-g, line 38: It would be nice to add some tests

Techassi: I actually thought about that. This would pull in at least `serde_json` or `serde_yaml` as a dev dependency. If we are fine with that, I can add a few unit tests.

core/src/serde.rs
```diff
@@ -0,0 +1,39 @@
+use proc_macro2::Span;
```
TedDriggs, line 1: I'd prefer that this impl be colocated with `IdentString`, rather than having a root module for serde functionality.

core/src/serde.rs
TedDriggs: I'm on the fence about this - I would like to see some tests, but I am worried about the impact on the build and test time.

core/src/serde.rs
```diff
@@ -0,0 +1,39 @@
...
+use syn::Ident;
+
+use crate::util::IdentString;
+
+impl<'de> serde::Deserialize<'de> for IdentString {
```
TedDriggs, line 6: Please add a doc comment somewhere that says this serializes as a string and expects a string.

core/src/serde.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(formatter, "a valid ident")
+    }
+
+    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        Ok(IdentString::new(Ident::new(v, Span::call_site())))
```
TedDriggs, line 28: This is going to panic if `v` isn't a valid identifier. Is there a way to make this return an error instead?

TedDriggs (changes requested):

core/src/serde.rs
Techassi: I can add some small tests and then we can see how the test times are impacted. If you are not satisfied, I can easily revert the commit.

core/src/serde.rs
Techassi: Alright, I can do that.

core/Cargo.toml
```diff
@@ -14,11 +14,13 @@ rust-version = "1.56"
 [features]
 diagnostics = []
 suggestions = ["strsim"]
+serde = ["dep:serde"]
```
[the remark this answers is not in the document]
Techassi: Oh wow, today I learned. Thank you! I will update it accordingly.

core/src/serde.rs
Techassi: Changed in 8dc6137.

core/Cargo.toml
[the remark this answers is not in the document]
Techassi: Removed in 888a14c.

core/src/serde.rs
Techassi: Added in e39a4bb.

Techassi:
Regarding the CI failures: We can either relax the requirements or drop back to an earlier version of `serde`. Let me know what path you prefer.

Techassi:
@TedDriggs were you able to take a look at this again?

Techassi:
Alright, makes sense. I will see how far serde needs to be downgraded to align with the current MSRV.

TedDriggs:
@Techassi if you want to take a look at this soon, I'm going to be running a release in the next few days.

Techassi:
Alright, will update this PR tomorrow.

Techassi:
This should now be mergeable.

TedDriggs (approved):

PR #346, opened 2025-06-04
Add `PreservedStrExpr` utility wrapper

m4tx:
By default, `darling` evaluates the expression inside string literals. This can be worked around with `preserve_str_literal`, but sometimes it's difficult—for instance, when a struct contains `HashMap<syn::Ident, syn::Expr>`.

This commit introduces a utility wrapper over `syn::Expr` that has a simple `FromMeta` impl that just copies the inner `Expr` instance. It can be then used like so:

```rust
#[derive(FromMeta)]
#[darling(attributes(demo))]
struct Demo {
    option: Option<HashMap<syn::Ident, PreservedStrExpr>>,
}
```

### DIFF
--- a/core/src/util/mod.rs
+++ b/core/src/util/mod.rs
@@ -12,6 +12,7 @@ mod path_to_string;
 mod shape;
 mod spanned_value;
 mod with_original;
+mod preserved_str_expr;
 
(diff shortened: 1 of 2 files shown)


TedDriggs:
This is very elegant.

PR #349, opened 2025-06-25
Add `syn::parse::Parse` implementation for `FromMeta` derive macro

### LINKED ISSUE #285: Implement `syn::Parse` for `FromMeta` types
I think it should be possible to codegen a syn::Parse implementation for structs that derive `From*` (I'm only looking at `FromMeta` because that's what I'm working with, but this can apply to other derive macros).

According to #62 a blanket implementation is not possible, but code generation might very well be, especially if it's opt-in. This works for me and emits errors just as expected:

```rust
#[derive(Debug, FromMeta)]
#[darling(syn_parse)]  // <-- attribute name open for bikeshedding
struct MyMacroArgs {
    // ...
}

// generated implementation:
impl Parse for MyMacroArgs {
    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
        let items =
            syn::punctuated::Punctuated::<NestedMeta, syn::Token![,]>::parse_terminated(input)?
                .into_iter()
                .collect::<Vec<_>>();
        Ok(MyMacroArgs::from_list(&items)?)
    }
}
```

This goes through a Vec because `FromMeta::from_list` takes a slice, but I guess that we could have a method that takes `impl IntoIterator` (or even implement `FromIterator`, maybe) to avoid collecting.

al8n:
Closes #285

With the `syn::parse::Parse`, when implement `#[proc_macro_attribute]`, the args parsing can be simplified to

```rust
use darling::{Error, FromMeta};
use darling::ast::NestedMeta;
use syn::ItemFn;
use proc_macro::TokenStream;

#[derive(Debug, FromMeta)]
struct MacroArgs {
    #[darling(default)]
    timeout_ms: Option<u16>,
    path: String,
}

#[proc_macro_attribute]
pub fn your_attr(args: TokenStream, input: TokenStream) -> TokenStream {
    let _args: MacroArgs = match syn::parse(args) {
        Ok(v) => v,
        Err(e) => { return e.to_compile_error().into(); }
    };
    let _input = syn::parse_macro_input!(input as ItemFn);

    // do things with `args`
    unimplemented!()
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ trybuild = "1.0.89"
 default = ["suggestions"]
 diagnostics = ["darling_core/diagnostics"]
 suggestions = ["darling_core/suggestions"]
+serde = ["darling_core/serde"]
 
 [workspace]
 members = ["macro", "core"]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -19,6 +19,10 @@ suggestions = ["strsim"]
 ident_case = "1.0.1"
 proc-macro2 = "1.0.86"
 quote = "1.0.18"
+serde = { version = "1.0.210", optional = true }
 syn = { version = "2.0.15", features = ["full", "extra-traits"] }
 fnv = "1.0.7"
 strsim = { version = "0.11.1", optional = true }
+
+[dev-dependencies]
+serde_json = "1.0.140"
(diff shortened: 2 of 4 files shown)


TedDriggs:
@al8n can you update your branch to be [here](https://github.com/TedDriggs/darling/tree/al8n/master)?

PR #352, opened 2025-06-25
Make `syn::TypePath` supports both quote-wrapped and bare values.

### LINKED ISSUE #351: `syn::TypePath` only supports both quote-wrapped values
I want to use bare values of `syn::TypePath` in my attributes, and got "Unexpected type `path`" error, which is surprising because similar types like `syn::Path` and `syn::ExprPath` support both quote-wrapped and bare values.

I can't replace `syn::TypePath` with any of them because:

- `syn::Path` does not accepts qualified path `<Foo as Bar>::Baz`.
- `syn::ExprPath` accepts extra attributes, it's error-prone to check and return an error if `expr_path.attrs.is_empty()`.

We can solve this by overriding `from_expr` method in `impl FromMeta for syn::TypePath`:

```rust
fn from_expr(expr: &syn::Expr) -> Result<Self> {
    match expr {
        syn::Expr::Path(body) => {
            if body.attrs.is_empty() {
                Ok(syn::TypePath {
                    qself: body.qself.clone(),
                    path: body.path.clone(),
                })
            } else {
                Err(Error::custom("attributes are not allowed").with_span(body))
            }
        }
        syn::Expr::Lit(expr_lit) => Self::from_value(&expr_lit.lit),
        syn::Expr::Group(group) => Self::from_expr(&group.expr),
        _ => Err(Error::unexpected_expr_type(expr)),
    }
}
```

JerryQ17:
Closes #351

### DIFF
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -474,7 +474,7 @@ impl FromMeta for syn::TypePath {
     /// Supports both quote-wrapped and bare values.
     fn from_expr(expr: &Expr) -> Result<Self> {
         match expr {
-            syn::Expr::Path(body) => {
+            Expr::Path(body) => {
                 if body.attrs.is_empty() {
                     Ok(syn::TypePath {
                         qself: body.qself.clone(),
@@ -484,8 +484,8 @@ impl FromMeta for syn::TypePath {
                     Err(Error::custom("attributes are not allowed").with_span(body))
                 }
             }
-            syn::Expr::Lit(expr_lit) => Self::from_value(&expr_lit.lit),
-            syn::Expr::Group(group) => Self::from_expr(&group.expr),
+            Expr::Lit(expr_lit) => Self::from_value(&expr_lit.lit),
+            Expr::Group(group) => Self::from_expr(&group.expr),
             _ => Err(Error::unexpected_expr_type(expr)),
         }
     }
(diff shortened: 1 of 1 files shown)


core/src/from_meta.rs
```diff
@@ -471,6 +470,39 @@ from_syn_parse!(syn::TypeTuple);
...
+            syn::Expr::Group(group) => Self::from_expr(&group.expr),
+            _ => Err(Error::unexpected_expr_type(expr)),
+        }
+    }
+
+    fn from_string(value: &str) -> Result<Self> {
+        syn::parse_str(value).map_err(|_| Error::unknown_value(value))
+    }
+
+    fn from_value(value: &::syn::Lit) -> Result<Self> {
```
TedDriggs, line 497: Nit: It looks like `syn::Lit` is already imported at the module level; we don't need this path here.

TedDriggs (approved):

JerryQ17:
I deleted those `syn::*` paths in `impl FromMeta for syn::TypePath` block.

I found other redundant `syn::Lit`, `syn::Expr` in this file, for example:

https://github.com/TedDriggs/darling/blob/39098b126e5ca69c30d6c6bb92bac8ccf2fe2175/core/src/from_meta.rs#L330-L337

Would you like to delete them all?

PR #354, opened 2025-07-01
impl `from_list` for `()`

### LINKED ISSUE #353: `()` should implement `from_list` to make the `flatten` work
Hi, may I ask to implement `from_list` for `()` to make the `flatten` work on `()` field? The following example will not work with the current implementation.

```rust
#[derive(FromMeta)]
struct MyMeta<M = ()> {
   foo: syn::Type,
   #[flatten]
   custom: M,
}
```

al8n:
Close #353.

### DIFF
--- a/core/src/from_meta.rs
+++ b/core/src/from_meta.rs
@@ -164,11 +164,21 @@ impl FromMeta for () {
     }
 
     fn from_list(items: &[NestedMeta]) -> Result<Self> {
-        if items.is_empty() {
-            Ok(())
-        } else {
-            Err(Error::too_many_items(0))
+        let mut errors = Error::accumulator();
+        for item in items {
+            errors.push(match item {
+                // Use `unknown_field_path` rather than `too_many_items` so that when this is used with
+                // `flatten` the resulting error message will include the valid fields rather than a confusing
+                // message about having only expected zero items.
+                //
+                // The accumulator is used to ensure all these errors are returned at once, rather than
+                // only producing an error on the first unexpected field in the flattened list.
+                NestedMeta::Meta(meta) => Error::unknown_field_path(meta.path()).with_span(meta),
+                NestedMeta::Lit(lit) => Error::unexpected_expr_type(lit.into()).with_span(lit),
+            });
         }
+
+        errors.finish()
     }
 }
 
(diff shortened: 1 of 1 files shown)


core/src/from_meta.rs
```diff
@@ -162,6 +162,14 @@ impl FromMeta for () {
     fn from_word() -> Result<Self> {
         Ok(())
     }
+
+    fn from_list(items: &[NestedMeta]) -> Result<Self> {
+        if items.is_empty() {
+            Ok(())
+        } else {
+            Err(Error::too_many_items(0))
+        }
```
TedDriggs, line 171: Given that the stated use-case for adding this method impl is to support `flatten`, this should accumulate all passed items as unknown field errors.

You can see in [`FlattenInitializer`](https://github.com/TedDriggs/darling/blob/master/core/src/codegen/field.rs#L133) how the error returned from this method is transformed to add the alternate keys if-and-only-if this returns an `ErrorKind::UnknownField` ([source](https://github.com/TedDriggs/darling/blob/aff17ba27f931ccf984eb55683f6f376ba77791d/core/src/error/mod.rs#L414)).

I think the rewrite of this is...

```suggestion
        let mut errors = Error::accumulator();
        for item in items {
            errors.push(match item {
                // Use `unknown_field_path` rather than `too_many_items` so that when this is used with
                // `flatten` the resulting error message will include the valid fields rather than a confusing
                // message about having only expected zero items.
                //
                // The accumulator is used to ensure all these errors are returned at once, rather than
                // only producing an error on the first unexpected field in the flattened list.
                NestedMeta::Meta(meta) => Error::unknown_field_path(meta.path()).with_span(meta),
                NestedMeta::Lit(lit) => Error::unexpected_expr_type(lit.into()).with_span(lit),
            });
        }

        errors.finish()
```

core/src/from_meta.rs
```diff
@@ -868,6 +876,13 @@ mod tests {
     #[test]
     fn unit_succeeds() {
         fm::<()>(quote!(ignore));
+        fm::<()>(quote!(ignore()));
+    }
+
+    #[test]
+    #[should_panic]
```
TedDriggs, line 883: Please include an `expected` on the `#[should_panic]` here.

TedDriggs (changes requested):
Directionally this is looking good.

Please add a test to `tests/flatten.rs` for this as well, showing that multiple forwarded items produce back multiple unknown field errors.

TedDriggs (commented):
A couple tweaks

PR #361, opened 2025-07-15
List valid fields on unknown field error

### LINKED ISSUE #325: UX improved error message idea: List valid attribute keys on error
First of, thanks for the crate. Lots of great stuff here. A co-worker suggested I take a look as I've been hand-rolling a lot of proc-macros recently. 

One of the things I've done is to list out all possible attribute keys on error, [implementation](https://github.com/schneems/magic_migrate/blob/b4a9bd4d88f3620ced3c243f8e65a1c70daae364/magic_migrate_derive/src/container_attributes.rs#L30-L34) and [another](https://github.com/heroku-buildpacks/cache_diff/blob/542f1fee58f2bdfb2dc05c3affc4b67e332191f3/cache_diff_derive/src/cache_diff_container.rs#L132-L138).

You can see the output of these implementations in [this test](https://github.com/heroku-buildpacks/cache_diff/blob/542f1fee58f2bdfb2dc05c3affc4b67e332191f3/cache_diff_derive/src/cache_diff_field.rs#L350-L351).

Hand-rolled output:

```
Unknown cache_diff attribute: `unknown`. Must be one of `rename`, `display`, `ignore`
```

Versus `darling::FromDeriveInput` output:

```
Unknown field: `unknown`
```

Through the suggestions feature I see errors like:

``` 
>Unknown field: `custommm`. Did you mean `custom`?`
```

So I'm guessing that there's a minimum threshold for the similarity score. When that's not met, instead of printing nothing, what if we printed all known keywords? Perhaps some people want to limit the number shown, it could be configurable, perhaps something like `#[darling(suggest(max_fallback = 8))]` to limit the number of attributes show to 8. Then (like the Rust compiler) it could emit `... and N more.` To indicate the list isn't exhaustive.

Alternatives: Always list out all valid on failure, even when there's a "did you mean" match.

TedDriggs:
Fixes #325


core/src/error/kind.rs
```diff
@@ -227,3 +255,23 @@ where
...
+    #[test]
+    fn present_no_alts() {
+        let err = ErrorUnknownField::new("hello", None);
+        assert_eq!(&err.to_string(), "Unknown field: `hello`");
+    }
+
+    #[test]
+    fn present_few_alts() {
+        let err = ErrorUnknownField::with_alts("hello", &["world", "friend"]);
+        eprintln!("{}", err);
```
TedDriggs, line 274:
```suggestion
```

Remove

PR #373, opened 2025-08-31
Remove fnv

DaniPopes:
`fnv` for use in `HashSet`s was added in #40, over 7 years ago, with no real justification.
This library is made for proc-macros where runtime performance doesn't matter as much, considering it will likely run in debug mode.

Remove `fnv` in favor of just using the standard library `HashSet` to remove a dependency.
This is a very small breaking change.


TedDriggs:
Won’t this change the public-facing type alias and be a breaking change?

DaniPopes:
Yes as mentioned in the PR!

Issue #219, opened 2023-01-09
Add support for assign default values to fields directly. eg. #[darling(default=false)]


RWDai:
nice crate! but i need some help.

if i want  assign true to bool by default,how can i do?

This is the implementation of my long-winded code now，how can i write it simply.
```
#[derive(FromMeta, Debug)]
struct MyMacors {
    #[darling(default)]
    _bool: FlipDeFaultBool,
}


#[derive(FromMeta,Debug)]
struct FlipDeFaultBool{
    _bool:bool
}

impl Default for FlipDeFaultBool {
    fn default() -> Self {
        FlipDeFaultBool{ _bool:true}
    }
}
```

if we can't do it better. Is there a way to add default values ​​in #[darling(default)] ?

juet like 

```
#[derive(FromMeta, Debug)]
struct MyMacors {
    #[darling(default=true)]
    _bool: bool,
}
```

thank you for you greate crate again!

TedDriggs:
The way to do this is to create a function that returns the value you want for the default and then pass the path as follows:

```rust
fn return_true() -> bool { true }

#[derive(FromMeta)]
struct Example {
  #[darling(default = "return_true")]
  some_field: bool,
}
```

I'm not looking to support other value types for defaults - the current implementation is simple and consistent, and works for both primitive and complex values.

cyqsimon:
> [@cyqsimon](https://github.com/cyqsimon) it might be possible to add support for:
> 
> `#[darling(default = || true)]` using the same technique as some of the other path-or-closure properties use.

I'll give it a shot

PR #376, opened 2025-09-19
Add support for closure as argument to `default`

cyqsimon:
- E.g. `#[darling(default = || get_val())]`
- See https://github.com/TedDriggs/darling/issues/219#issuecomment-3308315497

### DIFF
--- a/core/src/options/core.rs
+++ b/core/src/options/core.rs
@@ -150,9 +150,7 @@ impl ParseData for Core {
         match self.data {
             Data::Struct(Fields {
                 style: Style::Unit, ..
-            }) => {
-                panic!("Core::parse_field should not be called on unit")
-            }
+            }) => panic!("Core::parse_field should not be called on unit"),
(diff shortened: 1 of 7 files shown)


core/src/options/core.rs
```diff
@@ -145,7 +150,9 @@ impl ParseData for Core {
         match self.data {
             Data::Struct(Fields {
                 style: Style::Unit, ..
-            }) => panic!("Core::parse_field should not be called on unit"),
+            }) => {
+                panic!("Core::parse_field should not be called on unit")
+            }
```
cyqsimon, line 155: Weird. Rustfmt decided to make this change when ran without #377, but llater when I added the empty `rustfmt.toml` it refuses to change it back.

Then if I manually change it back, it somehow has no complaints with it either? Weird 😅

TedDriggs:
I will review this in the next few days.

Issue #340, opened 2025-05-14
feature request: support `darling(crate = "path/to/darling")` like `serde(crate = "path/to/serde")`


al8n:
Hi, thanks for the amazing crate!

I am developing an intermediate proc-macro crate that will generate structs or enums using darling. Without specifying the darling crate, the output will report `darling is not found` when the crates do not have the `darling` available at the root level.

TedDriggs:
Can you share more about how `serde` does this? I am willing to explore this, but fixing it must not cause complexity increases across the rest of the codebase.

al8n:
> Can you share more about how `serde` does this? I am willing to explore this, but fixing it must not cause complexity increases across the rest of the codebase.

For example, currently, the trait path is hard-coded to `::darling`

```
impl<'a> OuterFromImpl<'a> for FromMetaImpl<'a> {
    fn trait_path(&self) -> syn::Path {
        path!(::darling::FromMeta)
    }

    fn base(&'a self) -> &'a TraitImpl<'a> {
        &self.base
    }
}
```

What needs to be changed is that there should be a configuration that can change the path from `::darling` to `path/to/darling`.

TedDriggs:
I don't see a way to do that without a lot of noise in the code, unfortunately. There are 139 occurrences of `::darling` across 13 different codegen files (and 3 `options` module files). 

Passing it in would require updating every struct that contains such a usage, as well as updating every top-level options struct to accept and propagate the new path. That would likely require: 

1. Adding a new `root: Option<Path>` field to `TraitImpl`
2. Adding an accessor to `ExtractAttribute` and implementing it for all the implementors
3. Solving the numerous edge cases that emerge

One of the nastiest examples is `darling_core::codegen::DefaultExpression`. It would require adding a new field `root` to the `Trait` variant, and that in turn is going to require splitting the enum so there's one in `darling_core::options` that doesn't have that field (since we don't want to accept a custom root on a per-`DefaultExpression`-callsite basis).

I understand you're hard-blocked by this, so I'm willing to review a PR that implements the above. However, I do have significant concerns about the code-complexity impact, and I am worried that we don't have a good way of putting in tests to avoid regressions in the future.

al8n:
Hi @TedDriggs, I noticed in #285, a new field was added in the `FromMetaOptions` struct. Basically, I think this feature also just needs to add a new field in `*Options`, will you accept it? If so, I can file a PR.

TedDriggs:
[...]

PR #395, opened 2025-12-11
Allow `#[darling(crate = ...)]` to specify custom crate instead of `darling`

nik-rev:
This PR:

- Adds support for `#[darling(crate = ...)]` to change the path where we search for the `darling` crate
- Wraps all generated code inside of a `const` block with a custom `use #path as _darling` if `darling(crate = ...)` was specified, otherwise uses `extern crate darling as _darling`
- Replaces all uses of `::darling::` with `_darling::` inside the codegen crate

Closes https://github.com/TedDriggs/darling/issues/340


core/src/options/core.rs
```diff
@@ -51,6 +54,7 @@ impl Core {
             generics: di.generics.clone(),
             data: Data::try_empty_from(&di.data)?,
             default: Default::default(),
+            krate: None,
```
TedDriggs:
```suggestion
            krate: Default::default(),
```

Let's keep consistency with other initializers.

core/src/util/wrap_in_const.rs
```diff
@@ -0,0 +1,43 @@
...
+/// This is done to support users defining a custom location for the `darling` crate
+///
+/// If not wrapped in a `const` block, then multiple uses of `derive` on `darling`'s
+/// macros will create errors, because there may be multiple `extern crate darling as _darling`
+///
+/// # Arguments
+///
+/// - `tokens`: The trait implementations to wrap inside of `const` block
+/// - `krate`: Path to the darling crate, which defaults to `darling`
+pub fn wrap_in_const<T: ToTokens>(tokens: &T, krate: Option<&syn::Path>) -> TokenStream {
```
TedDriggs: This belongs in `codegen`, not `util` - it's not something that a `darling` consumer would directly use in its own source code.

nik-rev:
I've made the changes. some things to note:

- Two `const _: () {...}` blocks will be generated for `derive_syn_parse` using the approach of modifying `OuterFromImpl::wrap`
- I've added a test that `#[darling(derive_syn_parse)]` works with a custom `crate`

tests/compile-fail/custom_crate_omitted.rs
```diff
@@ -0,0 +1,9 @@
+mod darling {}
+
+#[derive(darling::FromDeriveInput)]
+// forgot to specify #[darling(crate = darling)]
+pub struct Example {
+    ident: Option<syn::Ident>,
+}
+
+fn main() {}
```
TedDriggs:
```suggestion
use ::darling as not_darling;

mod darling {}

#[derive(not_darling::FromDeriveInput)]
// forgot to specify #[darling(crate = not_darling)]
pub struct Example {
    ident: Option<syn::Ident>,
}

fn main() {}
```

As written, this test isn't showing what happens when the macro is actually invoked and tries to use the default `darling` path only to have that not work.

TedDriggs: Huh. Okay, in that case let's just delete this test.

Issue #396, opened 2025-12-12
Add `darling::util` functions for extracting value from an `Option<T>`


nik-rev:
Having macros that require hard-coded support for `Option<T>` is pretty common. I found many (130+) manual re-implementations of the same function (`fn(syn::Type) -> Option<syn::Type>`) which extracts the inner `syn::Type` from an `Option<T>`:

- https://github.com/search?q=path%3A*.rs+%2Ffn+.*option%5C%28.*%3A+.*syn%3A%3AType%5C%29+-%3E+Option%3C.*syn%3A%3AType%3E%2F&type=code

I think it would be useful for `darling` to include some helpers for extracting from an option.

Maybe 1 of these functions, or all 3:

```rust
/// Extracts type from an `Option<T>`
///
/// proc-macros have no type information, so it's impossible for us to tell
/// if something is truly an `Option`. We can only guess. If the path ends with
/// an `Option`, (like in `Option`, `::std::option::Option`, or
/// `option::Option`), then we consider it to be one.
fn extract_option(ty: syn::Type) -> Option<syn::Type> {
    if let syn::Type::Path(path) = ty
        && path.qself.is_none()
        && path.path.segments.last()?.ident == "Option"
        && let syn::PathArguments::AngleBracketed(params) =
            path.path.segments.first()?.arguments
        && let syn::GenericArgument::Type(ty) = params.args.first()?
    {
        Some(ty)
    } else {
        None
    }
}

fn extract_option_ref(ty: &syn::Type) -> Option<&syn::Type> {
    if let syn::Type::Path(path) = ty
        && path.qself.is_none()
        && path.path.segments.last()?.ident == "Option"
        && let syn::PathArguments::AngleBracketed(params) =
            &path.path.segments.first().as_ref()?.arguments
        && let syn::GenericArgument::Type(ty) = params.args.first()?
    {
        Some(ty)
    } else {
        None
    }
}

fn extract_option_mut(ty: &mut syn::Type) -> Option<&mut syn::Type> {
    if let syn::Type::Path(path) = ty
        && path.qself.is_none()
        && path.path.segments.last()?.ident == "Option"
        && let syn::PathArguments::AngleBracketed(params) =
            &mut path.path.segments.first().as_mut()?.arguments
        && let syn::GenericArgument::Type(ty) = params.args.first()?
    {
        Some(ty)
    } else {
        None
    }
}
```

PR #398, opened 2025-12-12
Add `darling::util::extract_option` helpers to extract `T` from `Option<T>`

nik-rev:
This PR creates `darling::util::extract_option` with 3 functions, for operating on `Type`, `&Type` and `&mut Type` 

```rust
pub fn from_owned(ty: Type) -> Result<syn::Type> 
```

I went with your idea, as well as having the functions return `Result` right away, so it can be nicely composed using the `?` operator

I considered using a macro to define the 3 functions, but I don't think it's worth it because people will use "goto definition"and these 3 functions are fairly simple so I think we can just copy-paste (also, each function is slightly different because of ownership requirements, so the macro would have a lot of edge cases)

Closes https://github.com/TedDriggs/darling/issues/396

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -24,5 +24,4 @@ syn = { version = "2.0.15", features = ["full", "extra-traits"] }
 strsim = { version = "0.11.1", optional = true }
 
 [dev-dependencies]
-darling = { path = ".." }
 serde_json = "1.0.140"
--- a/core/src/util/extract_option.rs
+++ b/core/src/util/extract_option.rs
@@ -32,6 +32,7 @@ use syn::{spanned::Spanned, Type};
 /// # Example
 ///
 /// ```
+/// # use darling_core as darling;
 /// use darling::util::extract_option;
 /// use quote::ToTokens;
 /// use syn::Type;
@@ -94,6 +95,7 @@ pub fn from_owned(ty: Type) -> Result<Type> {
 /// # Example
 ///
 /// ```
+/// # use darling_core as darling;
 /// use darling::util::extract_option;
 /// use quote::ToTokens;
 /// use syn::Type;
@@ -156,6 +158,7 @@ pub fn from_mut(ty: &mut Type) -> Result<&mut Type> {
 /// # Example
 ///
 /// ```
+/// # use darling_core as darling;
 /// use darling::util::extract_option;
 /// use quote::ToTokens;
 /// use syn::Type;


core/src/util/extract_option.rs
```diff
@@ -0,0 +1,209 @@
+//! Extract type `T` from an `Option<T>`
```
TedDriggs:
```suggestion
//! Functions to extract a type `T` from a type declaration `Option<T>`.
```

core/src/util/extract_option.rs
```diff
@@ -0,0 +1,209 @@
...
+//!
+//! Procedural macros often have to hard-code support for `Option`s, so much so that the logic
+//! has been re-implemented [hunreds](https://github.com/search?q=path%3A*.rs+%2Ffn+.*opt.*%5C%28.*%3A+.*syn%3A%3AType%5C%29+-%3E+Option%3C.*syn%3A%3AType%3E%2F&type=code) of times across the ecosystem.
+//!
+//! Considering this common need, darling provides 3 functions for extracting a [`Type`] from an `Option<T>`:
+//!
+//! - [`from_owned`]: `fn(Type) -> Result<Type>`
+//! - [`from_mut`]: `fn(&mut Type) -> Result<&mut Type>`
+//! - [`from_ref`]: `fn(&Type) -> Result<&Type>`
```
TedDriggs:
```suggestion
```

The justification is great for the GitHub issue, but unnecessary in the module documentation.

Listing the different signatures in the module-level docs is unnecessary, since on docs.rs the functions will be listed directly below.

core/src/util/extract_option.rs
```diff
@@ -0,0 +1,209 @@
...
+//!
+//! These functions will return `Ok` given that the given [`Type`] is a [`Type::Path`], with a single generic
+//! argument, and the last path segment is `"Option"`
+
+use crate::{Error, Result};
+use syn::{spanned::Spanned, Type};
+
+/// Extracts a `Type` `T` from an `Option<T>`
+///
+/// For more information, see the [module-level](self) documentation
```
TedDriggs: I think a runnable example would be good here.

core/Cargo.toml
```diff
@@ -24,4 +24,5 @@ syn = { version = "2.0.15", features = ["full", "extra-traits"] }
 strsim = { version = "0.11.1", optional = true }
 
 [dev-dependencies]
+darling = { path = ".." }
```
TedDriggs: Does this not create a circular dependency?

nik-rev: I've added a `# use darling_core as darling` in the examples and removed the dev-dependency.

TedDriggs (approved):

PR #404, opened 2025-12-27
Add `#[darling(transparent)]` to delegate impl to inner field even if it is named

nik-rev:
This PR adds `#[darling(transparent)]` which allows delegating trait implementation to the inner type (in a `struct` or `enum` variant) if the field is named. Previously, this only worked with unnamed fields.

```rust
#[derive(Debug, FromMeta, PartialEq, Eq)]
#[darling(transparent)]
struct Lorem {
    named: bool,
}
```


core/src/ast/data/mod.rs
```diff
@@ -106,6 +106,22 @@ impl<V, F> Data<V, F> {
...
 
+    /// Get the `Data`'s struct fields
+    pub fn as_struct(&self) -> Option<&Fields<F>> {
+        match self {
+            Data::Enum(_) => None,
+            Data::Struct(f) => Some(f),
+        }
+    }
+
+    /// Get the `Data`'s enum variants
```
TedDriggs:
```suggestion
    /// Returns the variants if `Data` is an enum.
```

Nit: Doc comments use the third-person singular for the verb.

core/src/options/from_attributes.rs
```diff
@@ -19,22 +22,22 @@ impl FromAttributesOptions {
...
         .parse_body(&di.data)?;
 
-        if !opts.is_newtype() && opts.base.attr_names.is_empty() {
+        let is_transparent = opts.base.container.data.as_struct().map(|fields| {
+            codegen::is_transparent(
+                fields.len(),
+                fields.style,
+                opts.base.container.transparent.is_present(),
+            )
+        }) == Some(true);
```
TedDriggs:
```suggestion
        }).unwrap_or(false);
```

core/src/codegen/transparent.rs
```diff
@@ -0,0 +1,49 @@
...
+use quote::{quote, ToTokens, TokenStreamExt};
+use syn::Ident;
+
+use crate::{
+    ast::{Fields, Style},
+    codegen::Field,
+};
+
+/// If this struct is "transparent"
+pub fn is_transparent(len: usize, style: Style, is_transparent: bool) -> bool {
```
TedDriggs: I would decompose this to have an `is_implicitly_transparent` function that looks at the size and style. The current signature is confusing to me - having an argument that is the same name as the function is odd.

core/src/codegen/transparent.rs
```diff
@@ -0,0 +1,49 @@
...
+        .expect("returned earlier on `.len() != 1`");
+
+    match fields.style {
+        Style::Tuple => Some((TransparentMember::Unnamed, field)),
+        Style::Struct if is_transparent => Some((TransparentMember::Named(field.ident), field)),
+        _ => None,
+    }
+}
+
+pub enum TransparentMember<'a> {
```
TedDriggs: Is there no `syn` type that allows for a named or unnamed member?

Issue #305, opened 2024-09-05
Using #[darling(with = ...)] should somehow work for fields whose types don't impl FromMeta


TedDriggs:
Right now, the following code doesn't work:

```rust
fn demo(meta: &syn::Meta) -> darling::Result<Vec<usize>> {
    // doesn't matter
}

#[derive(FromDeriveInput)]
pub struct Example {
    #[darling(with = demo)]
    field: Vec<usize>
}
```

This fails to compile because `Vec<usize>` does not impl `FromMeta`.

The problem is the generated code for handling a missing field, which attempts to call `<Vec<usize> as FromMeta>::from_none`; that doesn't exist, and so the code doesn't compile. Adding a `default` for the field or struct works fine, but that shouldn't be required to make this work.

nik-rev:
For anyone interested, this is the block of code that the issue is referencing

`codegen/field.rs:245`

```rust
/// Creates an error if a field has no value and no default.
pub struct CheckMissing<'a>(&'a Field<'a>);

impl ToTokens for CheckMissing<'_> {
    fn to_tokens(&self, tokens: &mut TokenStream) {
        if !self.0.multiple && self.0.default_expression.is_none() {
            let ident = self.0.ident;
            let ty = self.0.ty;
            let name_in_attr = &self.0.name_in_attr;

            // If `ty` does not impl FromMeta, the compiler error should point
            // at the offending type rather than at the derive-macro call site.
            let from_none_call =
                quote_spanned!(ty.span()=> <#ty as _darling::FromMeta>::from_none());

            tokens.append_all(quote! {
                if !#ident.0 {
                    match #from_none_call {
                        _darling::export::Some(__type_fallback) => {
                            #ident.1 = _darling::export::Some(__type_fallback);
                        }
                        _darling::export::None => {
                            __errors.push(_darling::Error::missing_field(#name_in_attr))
                        }
                    }
                }
            })
        }
    }
}
```

Issue #256, opened 2023-11-15
Allow usage of `map` to handle fields that don't impl FromMeta


danieleades:
i may be misunderstanding the use of the 'mapping' feature

I'm attempting to parse a PathBuf from a derive input. something like-

```rust
#[derive(Debug, FromDeriveInput)]
#[darling(attributes(my_attribute))]
struct Args {
    #[darling(map = PathBuf::from)]
    path: PathBuf,
}
```

to parse-

```rust
#[derive(MyTrait)]
#[my_attribute(path = "some/path")]
struct MyStruct;
```

is this how 'map' is supposed to be used?

Same question for parsing a comma-separated list into a `Vec<String>`

danieleades:
just to be clear, the following does not compile using darling v0.20.3

```rust
use std::path::PathBuf;

use darling::FromDeriveInput;


#[derive(Debug, FromDeriveInput)]
#[darling(attributes(my_attribute))]
struct Args {
    #[darling(map = PathBuf::from)]
    path: PathBuf,
}

```

and neither does this

```rust
use std::path::PathBuf;

use darling::FromDeriveInput;


#[derive(Debug, FromDeriveInput)]
#[darling(attributes(my_attribute))]
struct Args {
    #[darling(map = "parse_path_from_string")]
    path: PathBuf,
}

fn parse_path_from_string(s: String) -> PathBuf {
    PathBuf::from(s)
}
````

TedDriggs:
The issue you're running into is what happens when the caller doesn't specify `path`. In that situation, `darling` will either use the function set in the field-level `#[darling(default = ...)]`, or it will use that field from the struct-level default, or it will call `FromMeta::from_none` on the field's type - in this case, that's `PathBuf`, so it's looking for `<PathBuf as FromMeta>::from_none` and failing to find it because `PathBuf` doesn't implement `FromMeta`.

There are a couple options here:

1. Setting a `default` at the field level does the trick. Note that I did run into #257 while testing this.
2. We could add `impl FromMeta for PathBuf` - I don't see any reason not to do so, and that would maximally simplify your code.

Here's an example using `default` to prevent that `from_none` call being generated

```rust
#[derive(Debug, FromDeriveInput)]
#[darling(attributes(my_attribute))]
struct Args {
    #[darling(map = parse_path_from_string, default)]
    path: Option<PathBuf>,
}

fn parse_path_from_string(s: String) -> Option<PathBuf> {
    Some(PathBuf::from(s))
}
[...]

PR #406, opened 2025-12-28
Don't require types to implement `FromMeta` when unnecessary

nik-rev:
This now compiles, where it previously wouldn't because `field` did not implement `FromMeta`

```rust
fn demo(meta: &syn::Meta) -> darling::Result<Vec<usize>> {
    // doesn't matter
}

#[derive(FromDeriveInput)]
pub struct Example {
    #[darling(with = demo)]
    field: Vec<usize>
}
```

How it works is described in the doc comment on the `from_meta::spec` module

Closes https://github.com/TedDriggs/darling/issues/305
Closes https://github.com/TedDriggs/darling/issues/256

### DIFF
--- a/tests/with_no_impl_from_meta.rs
+++ b/tests/with_no_impl_from_meta.rs
@@ -15,7 +15,7 @@ fn parse_strings(_: String) -> Vec<String> {
     Vec::from(["x".to_string()])
 }
 
-#[derive(FromDeriveInput)]
+#[derive(FromDeriveInput, Debug)]
 #[darling(attributes(example))]
 pub struct Example {
     #[darling(with = demo)]
@@ -25,7 +25,7 @@ pub struct Example {
 }
 
 #[test]
-fn test() {
+fn pass() {
     let input = quote! {
         #[example(field, strings = "")]
         struct Example;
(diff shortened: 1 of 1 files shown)


core/src/lib.rs
```diff
@@ -36,6 +36,10 @@ pub use self::from_meta::FromMeta;
 pub use self::from_type_param::FromTypeParam;
 pub use self::from_variant::FromVariant;
 
+// Specialization via autoref trick
+#[doc(hidden)]
+pub use self::from_meta::spec as from_meta_spec;
```
TedDriggs: This is a small nit: To me "spec" sounds like it's a test. Is this naming common in other crates using the trick and I'm behind the times?

tests/with_no_impl_from_meta.rs
```diff
@@ -0,0 +1,37 @@
...
+fn test() {
+    let input = quote! {
+        #[example(field, strings = "")]
+        struct Example;
+    };
+    let input = syn::parse2::<DeriveInput>(input).unwrap();
+    let input = Example::from_derive_input(&input).unwrap();
+    assert_eq!(input.field, Vec::from([4]));
+    assert_eq!(input.strings, Vec::from(["x".to_string()]));
+}
```
TedDriggs: The really exciting thing here is that `field` is now required; can we add a second test that shows how omitting the field causes `Example::from_derive_input` to return an error?

For my own sanity, I tend to just check the count of errors rather than trying to build an exact copy of the error in the test.

TedDriggs (changes requested):
If it weren't for gen AI having ruined the use of emoji, I'd respond to this with 🤯. This is such a crazy trick to achieve a completely seamless experience for darling's direct consumers.

Minor changes requested, but then this will 100% be merging.

core/src/lib.rs
nik-rev: `Spec` is a common abbreviation for "Specialization" at least in the Rust compiler ([example](<https://github.com/rust-lang/rust/issues/89948>)) but since it's used there hundreds of times and here - just once, using the more verbose `autoref_specialization` will be clearer, especially since it's not a module exposed to users but it's just for the internal implementation of the macro

Issue #378, opened 2025-09-19
Cannot parse a raw visibilty specifier (e.g. `pub`, `pub(crate)`) to `syn::Visibility`


cyqsimon:
I'm trying to migrate [my pet project](https://github.com/cyqsimon/documented) from a bespoke macro argument parsing system to darling when I ran into this issue:

Currently my attribute macro allows the user to specify a custom visibility for the generated constant like so: `#[docs_const(vis = pub)]` ([docs](https://docs.rs/documented/0.9.2/documented/attr.docs_const.html#1-set-a-custom-constant-visibility-like-so)). So naturally for the migration I had something like this:

```rust
#[derive(darling::FromMeta)]
#[darling(derive_syn_parse)]
struct Args {
    vis: Option<Visibility>,
    // other options ...
}
```

But it seems like this won't allow `#[docs_const(vis = pub)]` (error: "expected an expression"); it will only accept `#[docs_const(vis = "pub")]`. Is there any way I can get my desired behaviour with what darling currently has?

cyqsimon:
I'll take a look tomorrow.

cyqsimon:
I concur with your view in general after doing some digging. I also found dtolnay/syn#1458 which I think may be what you are referring to.

From my understanding, what's giving us trouble is the fact that [`MetaNameValue.value`](https://docs.rs/syn/2.0.106/syn/struct.MetaNameValue.html) is an `Expr`, which means we would have to "build on top" if we want to support things like raw visibility. This is a rather unfortunate restriction on `syn`'s part, but if it's to align with the compiler implementation then I find it understandable.

The best idea I currently have is to extend [`NestedMeta`](https://docs.rs/darling/0.21.3/darling/ast/enum.NestedMeta.html) by adding an "escape hatch" (i.e. a [`TokenTree`](https://docs.rs/proc-macro2/1.0.101/proc_macro2/enum.TokenTree.html) variant), which catches anything that is neither a [`NestedMeta::Meta`] nor [`NestedMeta::Lit`]. Then also add the trait function `FromMeta::from_token_tree` with a default error implementation that implementers can override, exactly like many of the existing functions on `FromMeta`.

I'm pretty sure this will work, but it does lack in elegance. If you feel it's worth a try I can go ahead with writing up a prototype.

cyqsimon:
I too appreciate simplicity where possible. That being said, I have two main concerns with this approach:

[...]

PR #408, opened 2025-12-28
Pass tokens that failed to parse as `Expr` to `FromMeta::from_verbatim`

nik-rev:
This PR adds a new `FromMeta::from_invalid_expr` which accepts an arbitrary `TokenStream` as an input. This can be used to parse arbitrary code that fails to parse as a `syn::Expr`. For example, this failed to parse previously:

```rust
#[method(vis = pub(crate))]
```

You used to have to use quotation marks for that:

```rust
#[method(vis = "pub(crate)")]
```

Because `pub(crate)` is not a valid expression. Now, it will call the `from_invalid_expr` method, passing it the `TokenStream` corresponding to the tokens `pub(crate)` because we failed to parse `pub(crate)` as an expression

Also, `syn::Type`, `syn::Type*`, `syn::Visibility` and `syn::WhereClause` don't require quotation marks anymore

Closes https://github.com/TedDriggs/darling/issues/378

### DIFF
--- a/core/src/ast/data/nested_meta.rs
+++ b/core/src/ast/data/nested_meta.rs
@@ -2,7 +2,7 @@ use proc_macro2::{Delimiter, TokenStream, TokenTree};
 use quote::ToTokens;
 use syn::{
     ext::IdentExt,
-    parse::{discouraged::Speculative, ParseStream, Parser},
+    parse::{discouraged::Speculative, ParseStream, Parser, StepCursor},
     punctuated::Punctuated,
     token::{self, Brace, Bracket, Paren},
     Expr, ExprLit, Ident, Lit, MacroDelimiter, Meta, MetaList, MetaNameValue, Path, PathSegment,
@@ -38,13 +38,18 @@ fn parse_meta_path<'a>(input: ParseStream<'a>) -> syn::Result<Path> {
     })
 }
 
-fn parse_meta_after_path<'a>(path: Path, input: ParseStream<'a>) -> syn::Result<Meta> {
+fn parse_meta_after_path<'a>(path: Path, input: ParseStream<'a>) -> syn::Result<NestedMeta> {
     if input.peek(token::Paren) || input.peek(token::Bracket) || input.peek(token::Brace) {
-        parse_meta_list_after_path(path, input).map(Meta::List)
+        parse_meta_list_after_path(path, input)
+            .map(Meta::List)
+            .map(NestedMeta::Meta)
     } else if input.peek(Token![=]) {
-        parse_meta_name_value_after_path(path, input).map(Meta::NameValue)
+        Ok(match parse_meta_name_value_after_path(path, input)? {
+            MetaNameValueAnyRhs::ValidExpr(meta) => NestedMeta::Meta(Meta::NameValue(meta)),
+            MetaNameValueAnyRhs::InvalidExpr(meta) => NestedMeta::NameValueInvalidExpr(meta),
+        })
     } else {
-        Ok(Meta::Path(path))
+        Ok(NestedMeta::Meta(Meta::Path(path)))
     }
 }
 
@@ -72,10 +77,17 @@ fn parse_meta_list_after_path<'a>(path: Path, input: ParseStream<'a>) -> syn::Re
     })
 }
 
+enum MetaNameValueAnyRhs {
+    /// RHS after `=` is a valid [`syn::Expr`]
+    ValidExpr(MetaNameValue),
+    /// RHS after `=` is invalid [`syn::Expr`]
+    InvalidExpr(MetaNameValueInvalidExpr),
+}
+
 fn parse_meta_name_value_after_path<'a>(
     path: Path,
     input: ParseStream<'a>,
-) -> syn::Result<MetaNameValue> {
+) -> syn::Result<MetaNameValueAnyRhs> {
     let eq_token: Token![=] = input.parse()?;
     let ahead = input.fork();
     let lit: Option<Lit> = ahead.parse()?;
@@ -88,13 +100,59 @@ fn parse_meta_name_value_after_path<'a>(
     } else if input.peek(Token![#]) && input.peek2(token::Bracket) {
         return Err(input.error("unexpected attribute inside of attribute"));
     } else {
-        input.parse()?
+        // `input.parse()` advances the original parser, but we
+        // want to backtrack in case parsing into an `Expr` fails
+        let input_fork = input.fork();
+
+        match input.parse() {
+            Ok(expr) => expr,
+            // This isn't a valid expression, it might be something like `pub(in crate::module)`
+            // so we want to save that and let the user parse it (e.g. into a [`syn::Visibility`]).
+            //
+            // For more details, see docs of `darling::util::decode_if_verbatim`
+            Err(error) => {
+                fn eat_until_comma<'c>(
+                    cursor: StepCursor<'c, '_>,
+                ) -> syn::Result<(TokenStream, syn::buffer::Cursor<'c>)> {
+                    let mut rest = *cursor;
+                    let mut ts = TokenStream::new();
+                    while let Some((tt, next)) = rest.token_tree() {
+                        match tt {
+                            TokenTree::Punct(punct) if punct.as_char() == ',' => {
+                                break;
+                            }
+                            tt => {
+                                ts.extend([tt]);
+                                rest = next
+                            }
+                        }
+                    }
+                    Ok((ts, rest))
+                }
+
+                // Eat everything from the start of the attribute, until the ',' token
+                // We'll then parse this into user's custom implementation of `FromMeta::from_verbatim`,
+                // if it exists.
+                let verbatim_input = input_fork.step(eat_until_comma)?;
+
+                // Advance the original parser past the ',' token, so the next
+                // attributes are properly parsed
+                input.step(eat_until_comma)?;
+
+                return Ok(MetaNameValueAnyRhs::InvalidExpr(MetaNameValueInvalidExpr {
+                    path,
+                    eq_token,
+                    value: verbatim_input,
+                    error: error.into(),
+                }));
+            }
+        }
     };
-    Ok(MetaNameValue {
+    Ok(MetaNameValueAnyRhs::ValidExpr(MetaNameValue {
         path,
         eq_token,
         value,
-    })
+    }))
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


core/src/from_meta.rs
```diff
@@ -154,6 +156,22 @@ pub trait FromMeta: Sized {
...
+    /// ```ignore
+    /// #[demo(vis = pub(in crate), rename_all = "kebab-case")]
+    /// ```
+    ///
+    /// We'll first try to parse `pub(in crate)` as an expression, but it isn't valid,
+    /// so instead those tokens (up until the `,`, or end of attribute)
+    /// will be sent to the [`syn::Expr::Verbatim`] as a [`TokenStream`]
+    #[allow(unused_variables)]
+    fn from_verbatim(tokens: &TokenStream) -> Result<Self> {
+        Err(Error::unexpected_type("TokenStream"))
```
TedDriggs:
```suggestion
        Err(Error::unexpected_type("invalid expression"))
```

The `TokenStream` type is internal to proc macros, and shouldn't be visible to customers of `darling`-using crates.

tests/vis.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl FromMeta for Visibility {
+    fn from_expr(expr: &syn::Expr) -> darling::Result<Self> {
+        let syn::Expr::Verbatim(tokens) = expr else {
+            return Err(darling::Error::custom("expected visibility"));
+        };
+
+        Ok(Self(syn::parse2::<syn::Visibility>(tokens.clone())?))
+    }
+}
```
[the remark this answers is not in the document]
nik-rev: I added the `from_verbatim` function to all of these types:

```rust
from_syn_parse!(syn::Type);
from_syn_parse!(syn::TypeArray);
from_syn_parse!(syn::TypeBareFn);
from_syn_parse!(syn::TypeGroup);
from_syn_parse!(syn::TypeImplTrait);
from_syn_parse!(syn::TypeInfer);
from_syn_parse!(syn::TypeMacro);
from_syn_parse!(syn::TypeNever);
from_syn_parse!(syn::TypeParam);
from_syn_parse!(syn::TypeParen);
from_syn_parse!(syn::TypePtr);
from_syn_parse!(syn::TypeReference);
from_syn_parse!(syn::TypeSlice);
from_syn_parse!(syn::TypeTraitObject);
from_syn_parse!(syn::TypeTuple);
from_syn_parse!(syn::Visibility);
from_syn_parse!(syn::WhereClause);
```

The HTML tag example is a great idea

core/src/util/expr_verbatim.rs
```diff
@@ -0,0 +1,89 @@
...
+/// )
+/// ```
+///
+/// The resulting error `TokenStream` will need to be turned into a [`darling::Error`](Error) with
+/// [`Error::from_syn_rendered`] in order to be returned from [`FromMeta::from_verbatim`].
+///
+/// This function could do that, but it doesn't - because doing so requires cloning the entire error
+/// `TokenStream`, which is extra work that doesn't need to be done if you actually need `Verbatim`.
+/// For example if you are parsing into [`syn::Visibility`]
+pub fn decode(maybe_verbatim: &syn::Expr) -> Option<(&TokenStream, &TokenStream)> {
```
TedDriggs: Rather than having this return two `&TokenStream`s, I'd suggest having a struct, returning an `Option<???>` and then having a trait impl from that into `darling::Error` and accessors for the two values.

You might be able to define a struct called `LazyError` and then return `Option<darling::util::WithOriginal<LazyError<'a>, &'a Expr>>`?

core/src/from_meta.rs
```diff
@@ -154,6 +166,56 @@ pub trait FromMeta: Sized {
     fn from_bool(value: bool) -> Result<Self> {
         Err(Error::unexpected_type("bool"))
     }
+
+    /// Any other tokens that are not valid expressions
```
TedDriggs: This current design feels dangerous.

Up until now, parsing has not produced a `syn::Meta` that would _output_ different tokens than one parsed by `syn` itself. All previous discussions have been limited to changes around whether keywords are considered idents, which didn't affect output code.

This seems to change that; attempting to output the `Meta` could now produce tokens that are different than the input and that won't compile.

I fear this may push us to having a `darling::ast::Meta` so we can store this information in a safer way. That's a huge breaking change though.

TedDriggs (commented):
I have general concerns about this, and think we may need to discuss pros and cons of different approaches before continuing review of a specific PR.

core/src/from_meta.rs
TedDriggs: Given this code:

```rust
#[derive(FromField)]
#[darling(attributes(outer))]
struct Field {
    thing: Thing
}

#[derive(FromMeta)]
struct Thing {
    arg: usize,
    inner: Inner,
}

#[derive(FromMeta)]
struct Inner {
    other: syn::LitStr,
    vis: syn::Visibility
}
```

For the input:

```rust
#[outer(thing(
    arg = 4,
    inner(
        other = "Hello",
        vis = pub(crate)
    )
))]
```

If I'm understanding correctly, our version of parsing a `Meta` would use `syn::Expr::Verbatim` where `syn`'s parsing would've produced an error, but we won't retain that error initially.

Later on, when we call `Inner::from_meta`, that will see the value is a list and call `Inner::from_list` having just parsed the items into `NestedMeta`. This is the moment at which we produced the `syn::Error` that we actually hold onto.

**Question 1:** What function of `FromMeta` does that call, and what is the parameter it receives? I think we need the `NestedMetaVerbatimNameValue` struct so it can be the parameter type for this new `from_invalid_meta` method.

**Question 2:** In this example, there is no comma after the `vis`; will we stop consuming tokens in the right place?

nik-rev: > If I'm understanding correctly, our version of parsing a Meta would use syn::Expr::Verbatim where syn's parsing would've produced an error, but we won't retain that error initially.

**Right now**, we *do* retain the error as a `TokenStream` when parsing the `NestedMeta`, there's no way around that if we want good error messages. We first have `syn::Error`, then we call `error.into_token_stream()` to turn that into a bunch of `compile_error!("...")` invocations (the `TokenStream`)

With the approach I mentioned above (adding new variant to `NestedMeta`), we store the `syn::Error` directly in the field, without needing to do any kind of "encoding"

> Later on, when we call Inner::from_meta, that will see the value is a list and call Inner::from_list having just parsed the items into NestedMeta. This is the moment at which we produced the syn::Error that we actually hold onto.

Yes.

`Meta::List` contains this `TokenStream`:

```rust
(
    other = "Hello",
    vis = pub(crate)
)
```

Which is then parsed into the following 2 `NestedMeta`s:

```rust
NestedMeta::Meta(Meta::NameValue(MetaNameValue {
    path: parse_quote(other),
    value: parse_quote!("Hello"),
}))
NestedMeta::VerbatimNameValue(NestedMetaVerbatimNameValue {
    path: parse_quote!(vis),
    input: quote!(pub(crate)),
    error: <syn error from Expr::parse(input)>
})
```

> Question 1: What function of FromMeta does that call, and what is the parameter it receives? I think we need the NestedMetaVerbatimNameValue struct so it can be the parameter type for this new from_invalid_meta method.

I like the name `FromMeta::from_invalid_expr`, maybe we can call the variant `NestedMeta::InvalidNameValueExpr`, it can receive the `NestedMetaInvalidNameValueExpr` as a parameter:

```rust
struct NestedMetaInvalidNameValueExpr {
        path: Path,
        eq_token: Token![=],
        // the original input
        input: TokenStream,
        // error that resulted from trying to parse into an Expr
        error: syn::Error
}
```

We have `FromMeta::from_expr` that strictly deals with valid expressions, so `FromMeta::from_invalid_expr` will deal with invalid expressions

I believe this is better than `from_invalid_meta` or `from_invalid_meta_name_value` because the *name* part of the `NestedMeta` is valid (e.g. `name = `) it is the *value* part which is not (the RHS is invalid):

```rs
vis = pub(crate)
^^^^^ valid
      ^^^^^^^^^^ invalid
```

> Question 2: In this example, there is no comma after the vis; will we stop consuming tokens in the right place?

Yes

This is the logic responsible for that:

```rust
fn eat_until_comma<'c>(
    cursor: StepCursor<'c, '_>,
) -> syn::Result<(TokenStream, syn::buffer::Cursor<'c>)> {
    let mut rest = *cursor;
    let mut ts = TokenStream::new();
    while let Some((tt, next)) = rest.token_tree() {
        match tt {
            TokenTree::Punct(punct) if punct.as_char() == ',' => {
                break;
            }
            tt => {
                ts.extend([tt]);
                rest = next
            }
        }
    }
    Ok((ts, rest))
}
```

This eats every token until it encounters a comma (the `break`), or end of input (`while let` aborts)

TedDriggs: This is sounding good: Existing crate users won't need to make any changes to preserve current behavior, calling `ToTokens` will always produce back the input we got, and we can unblock things like types and visibilities in RHS positions.

core/src/ast/data/nested_meta.rs
```diff
@@ -103,8 +161,84 @@ fn parse_meta_name_value_after_path<'a>(
...
+/// corresponding to `pub(crate)` in this struct, as well as the original [`syn::Error`]
+/// obtained from <code>\<[syn::Expr] as [Parse](syn::parse::Parse)\>::parse(quote!(pub(crate)))</code>
+///
+/// This [`syn::Error`] is used for good error messages when a type does not implement
+/// [`FromMeta::from_invalid_expr`], as the default implementation of that function just the
+/// stored error.
+#[derive(Debug, Clone)]
+pub struct MetaNameValueInvalidExpr {
+    /// In `vis = pub(crate)`, this is the `vis`
+    path: Path,
```
TedDriggs: Why not make these fields `pub`?

core/src/from_meta.rs
nik-rev: I've just pushed a new iteration of the design. as discussed, it adds a new variant to `NestedMeta` without forcing us to resort to hacks.

core/src/error/mod.rs
```diff
@@ -310,6 +310,54 @@ impl Error {
...
         Default::default()
     }
+
+    /// Creates `Self` from the `TokenStream` obtained via `syn::Error::into_token_stream()`
+    ///
+    /// You might only need this if you overwrite the [`FromMeta::from_expr`](crate::FromMeta::from_expr)
+    /// method when implementing it.
+    ///
+    /// See docs in [`util::decode_if_verbatim`](crate::util::decode_if_verbatim) for more info
+    pub fn from_syn_rendered(syn_rendered: TokenStream) -> Self {
```
TedDriggs: Is this still used?

core/src/ast/data/nested_meta.rs
```diff
@@ -103,8 +161,87 @@ fn parse_meta_name_value_after_path<'a>(
...
+///
+/// This [`syn::Error`] is used for good error messages when a type does not implement
+/// [`FromMeta::from_invalid_expr`], as the default implementation of that function just
+/// returns the stored error.
+#[derive(Debug, Clone)]
+pub struct MetaNameValueInvalidExpr {
+    path: Path,
+    eq_token: Token![=],
+    value: TokenStream,
+    error: syn::Error,
```
TedDriggs: Should we eagerly convert this to a `darling::Error`?

core/src/from_meta.rs
```diff
@@ -1036,21 +1158,47 @@ impl FromMeta for HashSet<Path> {
...
+        assert_eq!(type1, type2, "6");
+        assert_eq!(f(t1), type1, "7");
+        assert_eq!(f(t2), type1, "8");
+    }
+
     /// parse a string as a syn::Meta instance.
     fn pm(tokens: TokenStream) -> ::std::result::Result<syn::Meta, String> {
         let attribute: syn::Attribute = parse_quote!(#[#tokens]);
         Ok(attribute.meta)
     }
```
TedDriggs, line 1191: The fact that we needed `pnm` makes me wonder what will happen if someone tries to use `parse_macro_input` on a `syn::DeriveInput` that is trying to use `#[foo(vis = pub(in thing))]`. However, `syn::DeriveInput` has a list of `syn::Attribute`, which in turn has this `meta` property.

As a result, `darling` wouldn't get an opportunity to use its `NestedMeta` at all. The recommended pattern for `darling` is:

```
let receiver = Receiver::from_derive_input(&parse_macro_input!(tokens));
```

That's going to emit a compile error and return.

What am I missing?

nik-rev: `examples/html.rs` works fine and uses the recommended pattern

```rust
fn main() {
    let input = quote::quote! {
        #[args(tag = <br />, foo = 10)]
        struct Input;
    };
    let input = syn::parse2::<syn::DeriveInput>(input).unwrap();
    let input = <Input as darling::FromDeriveInput>::from_derive_input(&input).unwrap();

    assert_eq!(
        input.tag,
        TagName(Ident::new("br", proc_macro2::Span::call_site()))
    );
    assert_eq!(input.foo, 10);
}
```

What gets populated in the `attributes` field is the `syn::Meta::List`, which contains a `TokenStream`. That `TokenStream` is `tag = <br />, foo = 10`. Then `darling` parses that into a bunch of `NestedMeta`s (separated by `,`)

TedDriggs: Ah, I'd forgotten about `syn::Meta::List`. I thought I was going crazy.

Issue #394, opened 2025-12-11
Regression in NestedMeta parsing for v0.23: Support for `::prefix` in Path removed, causing error with `#[derive(Debug, ::prost::Message)]`


echou:
In version 0.23, the parsing of `NestedMeta` no longer supports the format `#[derive(Debug, ::prost::Message)]`, resulting in the error: "expected identifier or literal".

The root cause is that the parser does not recognize `::` as a valid prefix for a Path.

This behavior is a regression, as version 0.20 correctly supported this logic. It is unclear why this functionality was removed in v0.23.

see 0.23:
https://github.com/TedDriggs/darling/blob/cfef4e58225fb72edeb712b70ab984bd59d3baf2/core/src/ast/data/nested_meta.rs#L136
0.20:
https://github.com/TedDriggs/darling/blob/82a51e0b65b158de02ffb5d753ea4cd03529743b/core/src/ast/data.rs#L434-L436

Steps to reproduce:​
Use a attribute macro with #[derive(Debug, ::prost::Message)]in code.
Attempt to parse it using NestedMetain v0.23.
Observe the error "expected identifier or literal".
Expected behavior:​

The parser should handle ::as a Path prefix without errors, as in v0.20.

Impact:​

This breaks compatibility with code that relies on absolute paths in attributes.

PR #409, opened 2025-12-28
Fix path starting with `::` not being recognized

nik-rev:
Closes https://github.com/TedDriggs/darling/issues/394


tests/path_prefix.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use darling::FromDeriveInput;
+use quote::quote;
+use syn::{parse_quote, DeriveInput, Path};
+
+#[derive(FromDeriveInput)]
+#[darling(attributes(derive))]
+struct Derive {
+    #[darling(flatten)]
+    derives: HashMap<Path, ()>,
```
TedDriggs: Can this be tested as a unit test of `nested_meta`? It doesn't feel like this requires a whole integration test.

PR #412, opened 2025-12-29
Add a `serde` example which re-implements Serde's attribute parsing with darling

nik-rev:
I think users will find it really beneficial to have a larger example of a darling re-implementation of an interface they may already be familar with - `serde`'s attributes

This example offers what no other example does - it represents the exact serde AST with maximal type safety, users will be able to use this as a starting point for their own more complex APIs.

### DIFF
--- a/examples/serde.rs
+++ b/examples/serde.rs
@@ -1,0 +1,453 @@
+//! Re-implementation of `serde`'s attribute parsing, with a maximally type-safe API
+#![allow(unused)]
+
+use darling::{
+    ast::{Data, Fields},
+    util::{Callable, Flag},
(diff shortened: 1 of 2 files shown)


examples/serde.rs
```diff
@@ -0,0 +1,462 @@
...
+                 deserialize,
+             }| Self::Each {
+                serialize,
+                deserialize,
+            },
+        )
+    }
+}
+
+impl FromDeriveInput for Input {
```
TedDriggs: Nit: This should be adjacent to `Input`. Knowing `darling` I expected that this was coming, but as a first-time reader I was confused why `Input` didn't have `FromDeriveInput` anywhere in sight.

Issue #410, opened 2025-12-28
Implement `FromMeta` for `Vec<Path>`, `HashSet<Path>`, `Vec<Ident>`, and `HashSet<Ident>`


nik-rev:
If you want to have something like `#[derive(A, B)]` then you can do it with `#[darling(flatten)]`:

```rust
use std::collections::HashMap;

use darling::FromDeriveInput;
use quote::quote;
use syn::{parse_quote, DeriveInput, Path};

#[derive(FromDeriveInput)]
#[darling(attributes(derive))]
struct Derive {
    #[darling(flatten)]
    derives: HashMap<Path, ()>,
}

#[test]
fn test() {
    let input = quote! {
        #[derive(Debug, ::prost::Message)]
        struct Input;
    };
    let input: DeriveInput = syn::parse2(input).unwrap();
    let input = Derive::from_derive_input(&input).unwrap();
    assert_eq!(
        input.derives,
        HashMap::from([
            (parse_quote!(Debug), ()),
            (parse_quote!(::prost::Message), ())
        ])
    );
}
```

But i'd prefer to have that `HashMap<Path, ()>` be either a `HashSet<Path>` or `Vec<Path>`, neither of which work.

martin-g:
> or `Vec<Set>`

You mean `Vec<Path>`, right ?

TedDriggs:
> Why not? Because HashMap<K, V> works with K being either String, Ident, or Path, I'd expect Vec<K> to work with those 3 as well. I'm not suggesting that anything else is supported, just those 3.

My thought process had been that `impl<T: FromMeta> FromMeta for Vec<T>` is ambiguous in terms of the input it will parse - `foo = [1, 2, 3]` or `foo(1, 2, 3)` both seem like reasonable things you might think `Vec<usize>` would accept. 

In contrast, `HashMap<String|Ident|Path, V: FromMeta>` is more clearly going to keep the `Meta` syntax going one layer deeper.

You're right that an impl which doesn't try to support all `T: FromMeta` avoids my biggest concern on this front.

> It's also easier to discover. Vec<Path> was the first thing I tried, whereas with PathList you have to know about it

That's a very good point.

The ones that seem most immediately useful are:

- `Vec<Ident>` and `HashSet<Ident>`
- `Vec<Path>` and `HashSet<Path>` 

I have a little heartburn about `String`; is it important to support that one now, or are you okay with omitting it?

nik-rev:
[...]

PR #415, opened 2026-01-04
Add `impl FromMeta` for `Vec<Ident>`, `Vec<Path>`, `HashSet<Ident>` and `HashSet<Path>`

nik-rev:
- Add `impl FromMeta` for `Vec<Ident>`, `Vec<Path>`, `HashSet<Ident>` and `HashSet<Path>`
- Add example of `#[derive]` implementation: `derive.rs`

Notes

- I considered implementing it with a macro like the `map!`, but the bodies are different and small enough that I don't think it'd be worth it
- `HashSet` errors on duplicates like the `HashMap` and `BTreeMap`, but `Vec` doesn't

Closes https://github.com/TedDriggs/darling/issues/410

### DIFF
--- a/examples/derive.rs
+++ b/examples/derive.rs
@@ -1,0 +1,35 @@
+//! Shows how to implement a `#[derive]`-like macro.
+
+use std::collections::HashSet;
+
+use darling::FromDeriveInput;
+use syn::{parse_quote, Path};
+
+#[derive(FromDeriveInput, PartialEq, Eq)]
+#[darling(attributes(derive))]
+pub struct Derive {
+    // Note that `impl FromMeta for HashSet<Path>` errors on duplicates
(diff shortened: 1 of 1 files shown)


core/src/util/path_list.rs
```diff
@@ -42,6 +43,7 @@ impl PathList {
     /// This function runs in `O(n * m)` time.
     /// It is believed that path lists are usually short enough that this is better than allocating a set containing the values
     /// of `other` or adding a conditional solution.
+    #[deprecated(note = "use `HashSet<Path>` instead", since = "TBD")]
```
TedDriggs: The behavior of `intersection` is very different from `HashSet<Path>` as a receiver, so I would not have this deprecation line at all.

PR #421, opened 2026-01-15
Convert errors in Accumulator::handle

### LINKED ISSUE #252: Convert errors in Accumulator::handle
Currently, `Accumulator::handle` accepts a `darling::Result<T>`. This doesn't play nice when using it with non-darling errors (for example, to handle `syn` error one has to write:

```rust
let parsed = accumulator.handle(syn::parse2(tokens).map_err(Into::into));
```

If `Accumulator::handle` wound accept `Result<T, E> where E: Into<darling::Error>`, it would allow to handle other errors without explicit conversion, the same way the `?` operator works

TedDriggs:
@nik-rev @DCNick3 review is welcome

This makes for slightly more concise use of Accumulator::handle by removing the need to do error conversion at the call site.

This is a breaking change for callers who were already calling into on the error path. Two places in generated code were broken as a result of the change, so the first commit fixes those to avoid the error.

Fixes #252

### DIFF
--- a/core/src/error/mod.rs
+++ b/core/src/error/mod.rs
@@ -828,12 +828,13 @@ impl Accumulator {
 
     /// Handles a possible error.
     ///
-    /// Returns a successful value as `Some`, or collects the error and returns `None`.
-    pub fn handle<T>(&mut self, result: Result<T>) -> Option<T> {
+    /// Returns a successful value as `Some`, or collects the error,
+    /// converts it to `darling::Error`, and returns `None`.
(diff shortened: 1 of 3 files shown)


core/src/error/mod.rs
```diff
@@ -828,12 +828,13 @@ impl Accumulator {
 
     /// Handles a possible error.
```
nik-rev, line 829: a more concise doc for this function could be:

```rust
/// Extracts the `Ok` value, otherwise handles the error
```

PR #423, opened 2026-02-06
Implement `ToTokens` for `SpannedValue`

### LINKED ISSUE #422: impl `ToTokens` for `SpannedValue`?
I just happened to have a `Fields<SpannedValue<T>>` where `T` also implements `ToTokens`, but that combination doesn't work when emitting tokens because `SpannedValue` doesn't implement `ToTokens`.

Is there a reason for that or can I push a quick PR adding it?

seritools:
Fixes #422

Let me know if there is more to do :)


TedDriggs:
@seritools can you rebase this so CI runs again on top of the syn v3 update?

seritools:
@TedDriggs done!

PR #436, opened 2026-08-18
Allow unused_qualifications in generated code

### LINKED ISSUE #435: Regression in 0.24: derived impls trigger `unused_qualifications` in downstream crates
Since 0.24.0, code generated by `FromDeriveInput` / `FromField` / `FromVariant` sets off rustc's `unused_qualifications` lint in the consuming crate. I checked 0.21.3 and 0.23.0 against the same input and both emit nothing, so this is new in 0.24.

What makes it unpleasant is that the warnings point at the user's own struct fields, and the suggestion ("remove the unnecessary path segments") removes zero characters. There's nothing to act on, and the natural first assumption is that you wrote something wrong.

Minimal repro using a proc-macro crate with `darling = "0.24"` and `syn = "3"`:

```rust
#![warn(unused_qualifications)]

use darling::{
    ast::{Data, Fields},
    FromDeriveInput, FromField, FromVariant,
};
use syn::{Expr, Generics, Ident, Type};

#[derive(FromDeriveInput)]
#[darling(attributes(serialize_size))]
struct Input {
    ident: Ident,
    generics: Generics,
    data: Data<Variant, Field>,
}

#[derive(FromField)]
#[darling(attributes(serialize_size))]
struct Field {
    ty: Type,
    seq_max_elems: Option<Expr>,
    fixed_size: Option<()>,
}

#[derive(FromVariant)]
#[darling(attributes(serialize_size))]
struct Variant {
    fields: Fields<Field>,
}
```

That produces five warnings, on lines 12 (`ident`), 14 (`data`), 21 and 22 (the two `Option`
fields), and 28 (`fields`) — all shaped like this:

```
warning: unnecessary qualification
  --> src/lib.rs:12:5
   |
12 |     ident: Ident,
   |     ^^^^^
```

I think the trigger is that through 0.23 the generated code used `::darling::export::…`, and rustc exempts `::`-rooted paths from this lint (rust-lang/rust#122435). 0.24 moved everything to the `_darling` alias, which isn't a global path, so the exemption stopped applying — and the magic-field codegen stamps the input's span on those paths, so the warnings surface on the user's fields.

One thing worth knowing for anyone who hits this: `#[allow(unused_qualifications)]` on the deriving struct does *not* silence it, since the generated impl is a sibling item inside the `const _` wrapper. It has to be a crate-level `#![allow(...)]`, which also suppresses genuine findings in your own code.

If an `allow` is an acceptable fix, this one line in `codegen/outer_from_impl.rs` does it:

```diff
             #[allow(clippy::manual_unwrap_or_default)]
             #[allow(clippy::needless_continue)]
+            #[allow(unused_qualifications)]
             impl #impl_generics #trayt for #ty_ident #ty_generics
```

I patched 0.24.0 locally with that and the count goes 5 → 0 with no other change in output. Happy to send it as a PR if you'd like

jsdanielh:
Through 0.23 the generated impls referred to items as `::darling::...`, which rustc exempts from the lint. The `_darling` alias is not a `::`-rooted path, so the warnings began surfacing on the consuming crate's own struct fields.

Fixes #435.


TedDriggs:
This will release in 0.24.1. I am holding off on releasing that until I better understand the regression on `#[darling(with)]`

PR #438, opened 2026-08-20
Don't require `FromMeta` for `with` fields given an invalid expression

### LINKED ISSUE #433: 0.24 regression, `#[darling(with = ...)]` requires `FromMeta` impl
This compiles on 0.23 but not on 0.24:

```rust
use darling::FromMeta;

#[derive(Default)]
struct NotFromMeta;

fn parser(_meta: &syn::Meta) -> darling::Result<NotFromMeta> {
    Ok(NotFromMeta)
}

#[derive(FromMeta)]
struct Opts {
    #[darling(default, with = parser)]
    field: NotFromMeta,
}
```

Output:
```
error[E0277]: the trait bound `NotFromMeta: darling::FromMeta` is not satisfied
  --> crates/sandbox/src/main.rs:12:31
   |
12 |     #[darling(default, with = parser)]
   |                               ^^^^^^ unsatisfied trait bound
   |
```

flisky:
dfa9f88 routed a name-value whose right-hand side fails to parse as a `syn::Expr` to `FromMeta::from_invalid_expr`, but emitted that call for every field -- including fields with `#[darling(with = ...)]`, whose type is not required to impl `FromMeta`. Such fields stopped compiling with `the trait bound `T: FromMeta` is not satisfied`.

Fixes #433


core/src/codegen/field.rs
```diff
@@ -162,6 +162,15 @@ impl ToTokens for MatchArm<'_> {
         let with_callable = &field.with_callable;
         let post_transform = field.post_transform.as_ref();
 
+        // A `with` callable only accepts a `syn::Meta`, and the field type is not required to
+        // impl `FromMeta`, so surface the stored parse error instead of going through
+        // `FromMeta::from_invalid_expr`.
+        let from_invalid_expr = match field.with_callable {
+            // `Borrowed` is a user-supplied `with`; `Owned` is our `from_meta` default.
+            Cow::Borrowed(_) => quote!(_darling::export::Err(__inner.error.clone())),
+            Cow::Owned(_) => quote!(_darling::FromMeta::from_invalid_expr(__inner)),
```
TedDriggs: I believe the compiletest failure is happening because now the trait resolution failure is happening in unspanned tokens, which is causing rustc to put the error on the `#[derive(...)]` instead of on the field's type.

```suggestion
            Cow::Borrowed(_) => quote_spanned!(with_callable.span() => _darling::export::Err(__inner.error.clone())),
            Cow::Owned(_) => quote_spanned!(with_callable.span() => _darling::FromMeta::from_invalid_expr(__inner)),
```

core/src/codegen/field.rs
```diff
@@ -162,6 +162,15 @@ impl ToTokens for MatchArm<'_> {
         let with_callable = &field.with_callable;
         let post_transform = field.post_transform.as_ref();
 
+        // A `with` callable only accepts a `syn::Meta`, and the field type is not required to
+        // impl `FromMeta`, so surface the stored parse error instead of going through
+        // `FromMeta::from_invalid_expr`.
+        let from_invalid_expr = match field.with_callable {
+            // `Borrowed` is a user-supplied `with`; `Owned` is our `from_meta` default.
```
TedDriggs: I'd be okay with shipping this as a 0.24.1 update to fix the regression, but relying on owned vs borrowed isn't a good way of differentiating these two cases long-term.

What are your thoughts on having a type (maybe just called `With`) that's something like...

```rust
enum With {
    Custom(Callable),
    Trait(Callable),
}

impl ToTokens for With {
    fn to_tokens(...) {
        // delegate to the variant
    }
}
```

core/src/codegen/field.rs
```diff
@@ -162,6 +162,15 @@ impl ToTokens for MatchArm<'_> {
         let with_callable = &field.with_callable;
         let post_transform = field.post_transform.as_ref();
 
+        // A `with` callable only accepts a `syn::Meta`, and the field type is not required to
+        // impl `FromMeta`, so surface the stored parse error instead of going through
+        // `FromMeta::from_invalid_expr`.
+        let from_invalid_expr = match field.with_callable {
```
TedDriggs:
```suggestion
        let from_invalid_expr = match with_callable {
```

Nit: We already have a local variable for `with_callable` in scope. We should either use it, or we should remove the local and just use `field.with_callable` throughout the function's body.

TedDriggs (changes requested):
I am extremely grateful to you for figuring this one out; I looked at this code and completely missed the `from_invalid_expr` implications.

TedDriggs:
@flisky here is a patch with the review comments addressed. I tried to push them, but didn't have access to your repo.

```
diff --git a/core/src/codegen/field.rs b/core/src/codegen/field.rs
index 0a33e95..205af7c 100644
--- a/core/src/codegen/field.rs
+++ b/core/src/codegen/field.rs
@@ -4,7 +4,7 @@ use proc_macro2::TokenStream;
 use quote::{quote, quote_spanned, ToTokens, TokenStreamExt};
 use syn::{spanned::Spanned, Ident, Type};
 
-use crate::codegen::{DefaultExpression, PostfixTransform};
+use crate::codegen::{DefaultExpression, OverrideableCallable, PostfixTransform};
 use crate::usage::{self, IdentRefSet, IdentSet, UsesTypeParams};
 
 /// Properties needed to generate code for a field in all the contexts
@@ -25,7 +25,7 @@ pub struct Field<'a> {
     /// An expression that will be wrapped in a call to [`core::convert::identity`] and
     /// then used for converting a provided value into the field value _before_ postfix
     /// transforms are called.
-    pub with_callable: Cow<'a, syn::Expr>,
+    pub with_callable: OverrideableCallable<'a>,
     pub post_transform: Option<&'a PostfixTransform>,
     pub skip: bool,
     pub multiple: bool,
@@ -163,12 +163,15 @@ impl ToTokens for MatchArm<'_> {
         let post_transform = field.post_transform.as_ref();
 
         // A `with` callable only accepts a `syn::Meta`, and the field type is not required to
-        // impl `FromMeta`, so surface the stored parse error instead of going through
-        // `FromMeta::from_invalid_expr`.
-        let from_invalid_expr = match field.with_callable {
-            // `Borrowed` is a user-supplied `with`; `Owned` is our `from_meta` default.
-            Cow::Borrowed(_) => quote!(_darling::export::Err(__inner.error.clone())),
-            Cow::Owned(_) => quote!(_darling::FromMeta::from_invalid_expr(__inner)),
+        // impl `FromMeta` when using a custom `with`, so surface the stored parse error instead
+        // of going through `FromMeta::from_invalid_expr` if the caller provided their own `with`.
+        let from_invalid_expr = match with_callable {
+            OverrideableCallable::Custom(_) => {
+                quote_spanned!(with_callable.span() => _darling::export::Err(__inner.error.clone()))
+            }
+            OverrideableCallable::Default(_) => {
+                quote_spanned!(with_callable.span() => _darling::FromMeta::from_invalid_expr(__inner))
+            }
         };
 
         // Errors include the location of the bad input, so we compute that here.
diff --git a/core/src/codegen/mod.rs b/core/src/codegen/mod.rs
index e7ffef4..406c03c 100644
--- a/core/src/codegen/mod.rs
+++ b/core/src/codegen/mod.rs
@@ -12,6 +12,7 @@ mod from_type_param;
 mod from_variant_impl;
 pub mod ident_field;
 mod outer_from_impl;
+mod overrideable_callable;
 mod postfix_transform;
 mod trait_impl;
 mod transparent;
@@ -31,6 +32,7 @@ pub use self::from_none::from_none_call;
 pub use self::from_type_param::FromTypeParamImpl;
 pub use self::from_variant_impl::FromVariantImpl;
 pub use self::outer_from_impl::OuterFromImpl;
+pub use self::overrideable_callable::OverrideableCallable;
 pub use self::postfix_transform::PostfixTransform;
 pub use self::trait_impl::TraitImpl;
 pub use self::transparent::extract_transparent;
diff --git a/core/src/options/input_field.rs b/core/src/options/input_field.rs
index f86a6a1..de172e1 100644
--- a/core/src/options/input_field.rs
+++ b/core/src/options/input_field.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 
 use syn::{parse_quote_spanned, spanned::Spanned};
 
-use crate::codegen;
+use crate::codegen::{self, OverrideableCallable};
 use crate::options::{Core, DefaultExpression, ParseAttribute};
 use crate::util::{Callable, Flag, SpannedValue};
 use crate::{Error, FromMeta, Result};
@@ -34,10 +34,15 @@ impl InputField {
                 .map_or_else(|| Cow::Owned(self.ident.to_string()), Cow::Borrowed),
             ty: &self.ty,
             default_expression: self.as_codegen_default(),
-            with_callable: self.with.as_ref().map(|w| w.as_ref()).map_or_else(
-                || Cow::Owned(parse_quote_spanned!(self.ty.span()=> _darling::FromMeta::from_meta)),
-                Cow::Borrowed,
-            ),
+            with_callable: self
+                .with
+                .as_ref()
+                .map(|w| OverrideableCallable::Custom(Cow::Borrowed(w)))
+                .unwrap_or_else(|| {
+                    OverrideableCallable::Default(Cow::Owned(
+                        parse_quote_spanned!(self.ty.span()=> _darling::FromMeta::from_meta),
+                    ))
+                }),
             skip: *self.skip.unwrap_or_default(),
             post_transform: self.post_transform.as_ref(),
             multiple: self.multiple.unwrap_or_default(),
```

flisky:
error[E0583]: file not found for module `overrideable_callable`
  --> core/src/codegen/mod.rs:15:1
   |
15 | mod overrideable_callable;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   
 Oops, it seems you forget `git add core/src/codegen/overrideable_callable.rs`:)
 
 ---
 
 I've applied your change without `overrideable_callable`, and invite you to my fork.
 
However, feel free to take it away:)

TedDriggs:
💀 oops. Let me recreate it.

PR #441, opened 2026-08-27
Allow `clippy::redundant_field_names` for generated impl blocks

### LINKED ISSUE #440: `clippy::redundant_field_names` when using `darling(multiple)`
```rs
#[derive(darling::FromMeta)]
pub struct Foo {
    #[darling(multiple)]
    pub bar: Vec<syn::LitStr>,
}
```

results in

```plain
warning: redundant field names in struct initialization
 --> src/lib.rs:4:9
  |
4 |     pub bar: Vec<syn::LitStr>,
  |         ^^^ help: replace it with: `bar`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_field_names
  = note: `#[warn(clippy::redundant_field_names)]` on by default
```

because the generated code ends in

```rs
_darling::export::Ok(Self { bar: bar })
```

Probably just needs an `#[allow(clippy::redundant_field_names)]` on the impl block.

seritools:
Fixes #440


TedDriggs:
Hopefully rust-lang/rust-clippy#5795 will at some point allow us to remove these; I don't love that we are having to add these when we've already told clippy that this code is auto-generated.

